PR #167, opened 2017-12-08
Data format

AlexeyKozhevin:
Fix bugs which appear when data_format is 'channels_first'
Rename all self functions into "get_...", static without "get_..."
Rename that functions in models
Fix problem with dtype of predicted labels (argmax returns int64)


dataset/models/tf/base.py
```diff
@@ -1320,3 +1187,213 @@ def se_block(cls, inputs, ratio, name='se', **kwargs):
             scale = tf.reshape(x, shape)
             x = inputs * scale
         return x
+
+    @staticmethod
```
roman-kh: why static? classmethod

PR #168, opened 2017-12-12
fix problems with channels_first

AlexeyKozhevin:
Fix problems with kwargs in upsample.

### DIFF
--- a/dataset/models/tf/base.py
+++ b/dataset/models/tf/base.py
@@ -617,6 +617,20 @@ class TFModel(BaseModel):
         channels_axis = self.channels_axis(tensor, **kwargs)
         return shape[channels_axis] if shape else None
 
+    @staticmethod
+    def get_num_channels(tensor, data_format='channels_last'):
+        """ Return number of channels in the input tensor
+        Parameters
+        ----------
+        tensor : tf.Tensor
+        Returns
+        -------
+        shape : tuple of ints
+        """
+        shape = tensor.get_shape().as_list()
+        axis = TFModel.channels_axis(data_format)
+        return shape[axis]
+
     def has_classes(self, tensor):
         """ Check if a tensor has classes defined in the config """
         config = self.get_tensor_config(tensor)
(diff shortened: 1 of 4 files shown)


dataset/models/tf/base.py
```diff
@@ -617,6 +617,20 @@ def num_channels(self, tensor, **kwargs):
         channels_axis = self.channels_axis(tensor, **kwargs)
         return shape[channels_axis] if shape else None
 
+    @staticmethod
```
roman-kh: why static? classmethod fits here, doesn't it?

dataset/models/tf/base.py
```diff
@@ -1166,7 +1180,8 @@ def _add_output_predictions(self, inputs, scope, attr_prefix, **kwargs):
 
     def _add_output_proba(self, inputs, scope, attr_prefix, **kwargs):
         _ = scope, kwargs
-        proba = tf.nn.softmax(inputs, name='predicted_proba')
+        axis = -1 if kwargs['data_format'] == 'channels_last' else 1
```
roman-kh: cls.channels_axis(tensor, **kwargs)

dataset/models/tf/linknet.py
```diff
@@ -124,7 +124,7 @@ def decoder_block(cls, inputs, filters, name, **kwargs):
         """
         upsample_args = cls.pop('upsample', kwargs)
 
-        num_filters = inputs.get_shape()[-1].value // 4
+        num_filters = cls.get_num_channels(inputs, kwargs.get('data_format')) // 4
```
roman-kh: why `get_`? there is a `num_channels` method.

AlexeyKozhevin: Static 'num_channels' exists in 'data_format' branch but in 'nn' num_channels isn't static method.

dataset/models/tf/base.py
AlexeyKozhevin: Yes, I forgot about it.

dataset/models/tf/base.py
AlexeyKozhevin: No, that method is used in linknet decoder_block so it must be static. In that branch I don't rename non-static methods into 'get_...'

AlexeyKozhevin: You're right.

PR #214, opened 2018-01-17
add config class

AlexeyKozhevin:


dataset/config.py
```diff
@@ -0,0 +1,182 @@
+""" Config class"""
+
+from copy import deepcopy
+
+class Config:
```
roman-kh: Make sure that it supports ** (double star) and looks as similar to `dict` as possible (__setitem__, __len__, __iter__, update, etc).

dataset/config.py
```diff
@@ -0,0 +1,182 @@
...
+            other = Config(other)
+        return Config({**self.flatten(), **other.flatten()})
+
+    def __radd__(self, other):
+        if isinstance(other, dict):
+            other = Config(other)
+        return other.__add__(self)
+
+    def __getitem__(self, ind):
+        return self.get(ind)
```
roman-kh: It should raise KeyError if there is no key.
Rename `ind` to `key`

PR #257, opened 2018-05-29
Fixes to work normally with dataset represented as a list. Some bugs with scipy_transforms were fixed

mikhailkin:
Please, review accurately https://github.com/mikhailkin/dataset/blob/31875e363ad2bdc8810db782eafb8d97ed183361/dataset/batch.py#L507
Tuple return looks strange, but maybe is used somewhere else.

**Probably, warnings on excess dimensions should be added.**

### DIFF
--- a/dataset/batch.py
+++ b/dataset/batch.py
@@ -1,15 +1,17 @@
 """ Contains basic Batch classes """
 
 import os
-import traceback
 import threading
+import traceback
 
 import dill
+
 try:
     import blosc
 except ImportError:
     pass
 import numpy as np
+
 try:
     import pandas as pd
 except ImportError:
@@ -36,7 +38,7 @@ class Batch(BaseBatch):
     components = None
 
     def __init__(self, index, preloaded=None, *args, **kwargs):
-        if  self.components is not None and not isinstance(self.components, tuple):
+        if self.components is not None and not isinstance(self.components, tuple):
             raise TypeError("components should be a tuple of strings with components names")
         super().__init__(index, *args, **kwargs)
         self._preloaded_lock = threading.Lock()
@@ -62,7 +64,6 @@ class Batch(BaseBatch):
                 self._local = threading.local()
             self._local.pipeline = val
         self._pipeline = val
-
 
     def deepcopy(self):
         """ Return a deep copy of the batch.
@@ -97,7 +98,6 @@ class Batch(BaseBatch):
         """ Create batch from another batch """
         return cls(batch.index, preloaded=batch._data)  # pylint: disable=protected-access
 
-
     @classmethod
     def merge(cls, batches, batch_size=None):
         """ Merge several batches to form a new batch of a given size
@@ -114,6 +114,7 @@ class Batch(BaseBatch):
         -------
         batch, rest : tuple of two batches
         """
+
         def _make_index(data):
             return DatasetIndex(np.arange(data.shape[0])) if data is not None and data.shape[0] > 0 else None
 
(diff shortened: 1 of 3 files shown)


dataset/pipeline.py
```diff
@@ -187,7 +187,7 @@ def __getattr__(self, name):
         if name[:2] == '__' and name[-2:] == '__':
             # if a magic method is not defined, throw an error
             raise AttributeError()
-        elif self._is_batch_method(name):
+        elif self._is_batch_method(name, cls=self.dataset.batch_class):
```
roman-kh: That would limit the search to children of `dataset.batch_class` which may not be desirable in many cases (e.g. an action of `radio.CTImagesBatch` can create an instance of `ImagesBatch` but the latter is not a child of the former.

dataset/batch.py
```diff
@@ -271,7 +272,7 @@ def __setstate__(self, state):
 
     @property
     def _empty_data(self):
-        return None if self.components is None else self._item_class()   # pylint: disable=not-callable
+        return None if self.components is None else self._item_class()  # pylint: disable=not-callable
```
roman-kh: Removing spaces is hardly necessary and should be avoided unless it clearly breaks PEP8.

dataset/batch.py
```diff
@@ -1,15 +1,17 @@
 """ Contains basic Batch classes """
 
 import os
-import traceback
 import threading
+import traceback
```
roman-kh: This is hardly needed here.

dataset/batch.py
```diff
@@ -1,15 +1,17 @@
...
 
 import dill
+
```
roman-kh: In fact this breaks PEP8 to some extent and, thus, should be avoided.

dataset/batch.py
```diff
@@ -386,12 +387,17 @@ def get_items(self, index, data=None, components=None):
...
             comps = components if components is not None else range(len(_data))
-            res = tuple(data_item[self.get_pos(data, comp, index)] if data_item is not None else None
-                        for comp, data_item in zip(comps, _data))
+            try:
+                res = tuple(data_item[self.get_pos(data, comp, index)] if data_item is not None else None
+                            for comp, data_item in zip(comps, _data))
+            except TypeError:
+                res = tuple(
+                    list(data_item[i] for i in self.get_pos(data, comp, index)) if data_item is not None else None
+                    for comp, data_item in zip(comps, _data))
```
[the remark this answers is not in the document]
mikhailkin: Yep, but inside code you work with batches as a lists, for e.g.:
https://github.com/analysiscenter/dataset/blob/master/dataset/decorators.py#L147
I don't quite understand what the reason to convert to array then back is.

dataset/batch.py
```diff
@@ -499,7 +505,8 @@ def apply_transform(self, ix, func, *args, src=None, dst=None, p=1., use_self=Fa
 
         if np.random.binomial(1, p):
             if use_self:
-                return func(self, *_args, **kwargs)
+                result = func(self, *_args, **kwargs)
+                return result  # if isinstance(result, tuple) else (result,)
```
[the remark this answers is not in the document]
mikhailkin: Yep, pardon me. I've tried to debug in a lazy way

dataset/pipeline.py
mikhailkin: I've got use-case where it crashed. Let me try to repeat

dataset/batch.py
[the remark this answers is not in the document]
roman-kh: Only some actions are wrapped in `inbatch_parallel` and only a fraction of those will use a post function which uses `all_results`.

The faster approach is to allocate an array for results in init function and then fill it in parallel invocations of an action. Thus the parallelized method may return just None as `all_results` is never used.

Despite all this, converting a list into an array is just shorter and perhaps faster than executing a separate branch with list comprehensions.

roman-kh:
Closing for inactivity. Feel free to reopen if needed.

PR #260, opened 2018-06-07
Get file name

dmylzenova:
get_file_name supports ct_batch
add components before loading (to make possible to get new components in self[ix])

### DIFF
--- a/dataset/batch.py
+++ b/dataset/batch.py
@@ -570,10 +570,29 @@ class Batch(BaseBatch):
         return self
 
     def _get_file_name(self, ix, src, write=False):
+        """ Get file name corresponding to current index.
+
+        Parameters
+        ----------
+        src : str or FilesIndex
+            if str, can be name of a file with extension in the indexed path (i.e self.index.dirs must be True)
+            if FilesIndex it must contain the same 
+
+
+            or if called from dump with write=True, src can be a name of a path where data
+            will be dumped in files named the same names as in current index (built from files)
+
+        """
         if isinstance(self.index, FilesIndex):
+            
+
+
+            
             if isinstance(src, str):
                 if write and not self.index.dirs:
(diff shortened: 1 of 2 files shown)


dataset/batch.py
```diff
@@ -568,19 +569,49 @@ def apply_transform_all(self, func, *args, src=None, dst=None, p=1., use_self=Fa
...
+            if FilesIndex it must contain the same indices values as in the self.index.
+
+        write : bool
+            write = True is designed to be called from dump. Then src must be str and stand for a name of a folder
+            where data will be dumped in files named the same names as in current index.
+        """
+        if isinstance(self.index, FilesIndex):
+            if isinstance(src, str):
+                if write and not self.index.dirs:
+                    file_name = self.index.get_fullpath(ix).replace('\\', '/').split('/')[-1]
```
roman-kh: There is `os.path.dirname`

dataset/batch.py
```diff
@@ -568,19 +569,49 @@ def apply_transform_all(self, func, *args, src=None, dst=None, p=1., use_self=Fa
...
+                    if not os.path.exists(src):
+                        os.makedirs(src)
```
roman-kh: Avoid side effects. This function should return a name without affecting a file system

dataset/batch.py
```diff
@@ -568,19 +569,49 @@ def apply_transform_all(self, func, *args, src=None, dst=None, p=1., use_self=Fa
...
+    def _get_file_name(self, ix, src, write=False):
+        """ Get file name corresponding to current index.
+
+        Parameters
+        ----------
+        src : str or FilesIndex
+            if str, can be name of a file with extension in the indexed path (i.e self.index.dirs must be True)
+            if FilesIndex it must contain the same indices values as in the self.index.
+
+        write : bool
```
roman-kh: This is not needed. The behaviour should be similar to all cases.

PR #263, opened 2018-07-14
Iou

AlexeyKozhevin:


roman-kh:
There is going to be a lot of changes in the `metrics` branch (and `iou` is already implemented there as well). Have patience.

PR #289, opened 2019-02-08
fix channels

SergeyTsimfer:
One change in base.py to fix `channels_first` behavior.
Moved `build_config()` methods to `Inception` base class.


roman-kh:
Please, make tests for all ``data_format`` options:
- when it's set in input config
- in common
- for single input
- for multiple inputs
etc

SergeyTsimfer:
Currently, some tests are failing to pass during Shippable check, while local pytest indicates that everything is fine. Each of the failing tests is trying to use 'channels_first' format and raises `Conv2DCustomBackpropFilterOp only supports NHWC` error.
I was able to reproduce same behavior by using old tensorflow package (version==1.8), and found out that HCWH format was not supported back then. Switching that check to newer tensorflow should solve the problem.

One more thing: in the last commit I use wildcard import in `conftest.py` to get all of the needed fixtures  located in `batchflow/fixtures/dataformat.py`. Better solution to this problem is to use plugin as I did in [earlier verion,](https://github.com/analysiscenter/batchflow/blob/4713697855589b83a08fe82193273d796111eb65/batchflow/conftest.py) but that requires executing `pytest ./batchflow/tests` command instead of just calling `pytest`.

batchflow/tests/dataformat_test.py
```diff
@@ -0,0 +1,187 @@
...
+# pylint: disable=import-error, wrong-import-position, no-name-in-module
+# pylint: disable=missing-docstring, redefined-outer-name
+import os
+import warnings
+warnings.filterwarnings("ignore")
+os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'
+
+import pytest
+import numpy as np
+import tensorflow as tf
```
roman-kh: You don't seem to use tf anywhere in your code

batchflow/tests/dataformat_test.py
```diff
@@ -0,0 +1,187 @@
...
+LOCATIONS = set(['initial_block', 'body', 'block', 'head'])
+
+
+@pytest.fixture()
+def model_setup():
+    """ Pytest fixture to generate fake dataset and model config with desired format.
+
+    Parameters
+    ----------
+    d_f: {'channels_last', 'channels_first'}
```
roman-kh: Why not `data_format`?

batchflow/tests/dataformat_test.py
```diff
@@ -0,0 +1,187 @@
...
+    Returns
+    -------
+    Pipeline
+        Test pipeline that consists of initialization of model and
+        preparing for training with given config.
+    """
+    def _get_model_pipeline(model_class, current_config):
+        test_pipeline = (Pipeline()
+                         .init_variable('current_loss')
+                         .init_model('dynamic', model_class,
```
roman-kh:
```suggestion
                         .init_model('dynamic', C('model_class')),
```

batchflow/tests/dataformat_test.py
```diff
@@ -0,0 +1,187 @@
...
+                                     'TestModel', current_config)
```
roman-kh:
```suggestion
                                     'TestModel', C('model_config'))
```

batchflow/tests/dataformat_test.py
```diff
@@ -0,0 +1,187 @@
...
+                         .to_array()
+                         .train_model('TestModel',
+                                      fetches='loss',
+                                      images=B('images'),
+                                      labels=B('labels'),
+                                      save_to=V('current_loss'))
+                         )
+        return test_pipeline
+    return _get_model_pipeline
```
roman-kh: You don't need a complicated fixture here. You can just return `Pipeline` and within a test function apply a config to it through `set_config(pipeline_config)`. For that to work, `pipeline_config` should contain keys `model_class` and `model_config`.

PR #292, opened 2019-02-26
Parametrize locations for data format

roman-kh:


batchflow/tests/dataformat_test.py
```diff
@@ -140,41 +139,36 @@ def test_block_dataformat(self, model_and_config):
 
 @pytest.mark.parametrize('model', AVAILABLE_MODELS)
 class Test_models:
```
SergeyTsimfer: Do we still need to use 'class' for single test?

batchflow/tests/dataformat_test.py
```diff
@@ -140,41 +139,36 @@ def test_block_dataformat(self, model_and_config):
...
+                              pytest.param('channels_first', marks=pytest.mark.xfail),
+                              'channels_last'])
+    def test_data_format(self, model, model_setup, pipeline, location, data_format):
+        """ We can explicitly pass 'data_format' to inputs or common
+
+        Notes
+        -----
+        If `data_format` is None, use a default value.
+
+        `channels_first` does not work in TF 1.12 on CPU as a corresponding pooling operation is not implemented yet.
```
SergeyTsimfer: Mine actually works fine with python 3.6.8 + TF 1.12.0. 
'Default MaxPoolingOp only supports NHWC on device type CPU' is raised when I run this test locally with python 3.7.1 and TF  1.13.0rc2, and same error is raised when check by Shippable is performed.

batchflow/tests/dataformat_test.py
```diff
@@ -140,41 +139,36 @@ def test_block_dataformat(self, model_and_config):
...
+        expected_data_format = data_format or 'channels_last'
+        dataset, model_config = model_setup(data_format=expected_data_format)
+        if data_format:
+            model_config[location + '/data_format'] = data_format
+        config = {'model_class': model, 'model_config': model_config}
+        test_pipeline = (pipeline << dataset).set_config(config)
+        batch = test_pipeline.next_batch(2, n_epochs=None)
+        model = test_pipeline.get_model_by_name('model')
+
+        assert model.data_format('images') == expected_data_format
```
SergeyTsimfer:
```suggestion
         if location == 'inputs/images':
            assert model.data_format('images') == expected_data_format
```
Otherwise it unnecessarily fails when 'channels_first' is passed through 'common'.

PR #297, opened 2019-03-18
Modify apply transform

dpodvyaznikov:
Now if `src` argument of `apply_transform`  is a list, it applies func to each element of `src` identically and independently.


batchflow/batch.py
```diff
@@ -493,22 +494,69 @@ def apply_transform(self, ix, func, *args, src=None, dst=None, p=None, use_self=
...
+
+            apply_transform(make_masks_fn, src='images', dst='masks')
+            apply_transform(apply_mask, src=('images', 'masks'), dst='images', use_self=True)
+            apply_transform_all(rotate, src=['images', 'masks'], dst=['images', 'masks'], p=.2)
         """
+        to_act = bool(p is None or np.random.binomial(1, p))
+
+        if isinstance(src, list) and len(src) == len(dst):
+            return tuple([self._apply_transform(ix, func, to_act, *args, src=item, use_self=use_self, **kwargs)
+                          for item in src])
```
akoryagin: `component` would look much better than `item`

dmylzenova: doesn't enter this condition if `dst` is not provided. `dst` must be equal to `src` if not provided.

dmylzenova: also doesn't work if src is a list of length 1.

dpodvyaznikov: And this condition is passed if `src` is a list, `dst` is str and `len(src)` happens to be equal to `len(dst)`.

batchflow/batch.py
```diff
@@ -493,22 +494,73 @@ def apply_transform(self, ix, func, *args, src=None, dst=None, p=None, use_self=
...
+
+    def _apply_transform(self, ix, func, to_act, *args, src=None, dst=None, use_self=False, **kwargs):
+        """ Apply a function to each item in the batch.
+
+        Parameters
+        ----------
+        func : callable
+            a function to apply to each item from the source
+
+        to_act: bool
```
roman-kh: There is no point in changing method signature. Leave `p` as in `apply_transform`

batchflow/batch.py
```diff
@@ -493,22 +494,73 @@ def apply_transform(self, ix, func, *args, src=None, dst=None, p=None, use_self=
...
+        Examples
+        --------
+
+        ::
+
+            apply_transform(make_masks_fn, src='images', dst='masks')
+            apply_transform(apply_mask, src=('images', 'masks'), dst='images', use_self=True)
+            apply_transform_all(rotate, src=['images', 'masks'], dst=['images', 'masks'], p=.2)
         """
+        to_act = bool(p is None or np.random.binomial(1, p))
```
roman-kh: You may set `p` just to 0 or 1.

batchflow/batch.py
```diff
@@ -493,22 +494,73 @@ def apply_transform(self, ix, func, *args, src=None, dst=None, p=None, use_self=
...
+
+        if isinstance(src, list) and len(src) > 1 and isinstance(dst, (list, tuple)) and len(src) == len(dst):
```
roman-kh: if `dst` is None, you won't get into this branch.

roman-kh: No need to expect `dst` to be a tuple here either.
If `src` is a list, let `dst` be a list too.

batchflow/batch.py
dmylzenova: line 560 in _apply_transform src cannot be list here because you checked it in apply_transform, if i am not mistaken

batchflow/batch.py
dpodvyaznikov: Changed it back

batchflow/batch.py
dpodvyaznikov: Done

batchflow/batch.py
dpodvyaznikov: Fixed, if `dst` is `None`, it is set to src

dpodvyaznikov: Done

roman-kh:
Make some tests to check if `apply_transform` works properly.

roman-kh:
What about tests?

batchflow/batch.py
roman-kh: Remove it from a docstring too

batchflow/tests/apply_transform_test.py
```diff
@@ -0,0 +1,96 @@
...
+    batch = (dataset.next_batch(BATCH_SIZE)
+             .load(src=DATA, dst='comp1')
+             .load(src=DATA, dst='comp2')
+            )
+    return batch
+
+
+@pytest.mark.parametrize('src', SINGLE_CASE)
+@pytest.mark.parametrize('dst,expectation', list(zip(SINGLE_CASE, EXPECTATION)))
+def test_one_to_one(src, dst, expectation, batch):
```
roman-kh: I can hardly understand what these tests do.
Simplify and clarify.

PR #308, opened 2019-05-16
Encoder decoder

SergeyTsimfer:
Modified `EncoderDecoder` model in order to:

- add ability to create encoder from individual blocks (e.g. `DenseNet.block`)

- same for embedding and decoder: now we can create them from individual blocks

- add parameter to control skip connections between corresponding encodings/decodings (`UNet`-like)

Also, arguments for upsampling layers are moved to separate key in order to prevent collisions; same for downsampling layers (when encoder is created from blocks). 

Changed order of layers in body of `DenseNet`: from `block-transition-block-transition-block-block` to `block-block-transition-block-transition-block` in order to stay consistent with `ResNet`. Added `make_encoder` method.

In `make_encoder` method of `ResNet` added error handling: otherwise error-messages are very confusing. 

Added tests for encoder-decoder model.

Examples of model configurations can be found either in docstring of class, or in the file with tests.

### DIFF
--- a/batchflow/models/tf/encoder_decoder.py
+++ b/batchflow/models/tf/encoder_decoder.py
@@ -17,7 +17,7 @@ class EncoderDecoder(TFModel):
 
     body : dict
         encoder : dict
-            base_class : TFModel
+            base : TFModel
                 Model implementing ``make_encoder`` method which returns tensors
                 with encoded representation of the inputs.
 
@@ -52,7 +52,7 @@ class EncoderDecoder(TFModel):
                 If int, the total upsampling factor for all stages combined.
                 If list, upsampling factors for each stage.
 
-            bridges : bool
+            skip : bool
                 Whether to concatenate upsampled tensor with stored pre-downsample encoding.
             upsample : dict
                 Parameters for upsampling (see :func:`~.layers.upsample`).
@@ -74,7 +74,7 @@ class EncoderDecoder(TFModel):
             'inputs': dict(images={'shape': B('image_shape')},
                            masks={'name': 'targets', 'shape': B('mask_shape')}),
             'initial_block/inputs': 'images',
-            'body/encoder': {'base_class': ResNet,
+            'body/encoder': {'base': ResNet,
                              'num_blocks': [2, 3, 4]
                              'filters': [16, 32, 128]},
             'body/embedding': {'layout': 'cna', 'filters': 256},
@@ -104,18 +104,18 @@ class EncoderDecoder(TFModel):
 
     Notes
     -----
-    When `base_class` is used for decoder creation, downsampling is done one less time than
+    When `base` is used for decoder creation, downsampling is done one less time than
     the length of `filters` (or other size-defining parameter) list in the `encoder` configuration.
     That is due to the fact that the first block is used as preprocessing of input tensors.
     """
     @classmethod
     def default_config(cls):
         config = TFModel.default_config()
-        config['body']['encoder'] = dict(base_class=None, num_stages=None, blocks=None)
-        config['body']['encoder']['downsample'] = dict(layout='p', pool_size=2, pool_strides=2)
-        config['body']['embedding'] = dict(base=None)
-        config['body']['decoder'] = dict(bridges=True, num_stages=None, factor=None, blocks=None)
-        config['body']['decoder']['upsample'] = dict(layout='tna')
+        config['body/encoder'] = dict(base=None, num_stages=None, blocks=None)
+        config['body/encoder/downsample'] = dict(layout='p', pool_size=2, pool_strides=2)
+        config['body/embedding'] = dict(base=None)
+        config['body/decoder'] = dict(skip=True, num_stages=None, factor=None, blocks=None)
+        config['body/decoder/upsample'] = dict(layout='tna')
         config['head'] = dict(layout='c', kernel_size=1)
         return config
 
@@ -149,7 +149,7 @@ class EncoderDecoder(TFModel):
 
     @classmethod
     def head(cls, inputs, targets, name='head', **kwargs):
-        """ Linear convolutions with kernel size of 1. """
+        """ Linear convolutions. """
         kwargs = cls.fill_params('head', **kwargs)
         with tf.variable_scope(name):
             x = cls.crop(inputs, targets, kwargs['data_format'])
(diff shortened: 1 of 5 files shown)


batchflow/models/tf/encoder_decoder.py
```diff
@@ -4,173 +4,288 @@
...
 
+    Parameters
+    ----------
     inputs : dict
-        dict with 'images' (see :meth:`~.TFModel._make_inputs`)
+        Dictionary with 'images' (see :meth:`~.TFModel._make_inputs`)
 
     body : dict
         encoder : dict
             base_class : TFModel
```
roman-kh: Let's call it `base`.

batchflow/models/tf/encoder_decoder.py
```diff
@@ -4,173 +4,288 @@
...
+
             factor : int or list of int
-                if int, the total upsampling factor for all blocks combined.
-                if list, upsampling factors for each block.
-            layout : str
-                upsampling method (see :func:`~.layers.upsample`)
+                If int, the total upsampling factor for all stages combined.
+                If list, upsampling factors for each stage.
+
+            bridges : bool
```
roman-kh: While people all over the world call them `skip`, we call them `bridges`. It doesn't seem right, does it?

batchflow/models/tf/encoder_decoder.py
```diff
@@ -4,173 +4,288 @@
...
+    That is due to the fact that the first block is used as preprocessing of input tensors.
     """
     @classmethod
     def default_config(cls):
         config = TFModel.default_config()
-        config['body']['encoder'] = dict(base_class=ResNet18)
-        config['body']['decoder'] = dict(layout='tna', factor=8, num_stages=3)
-        config['body']['embedding'] = dict(layout='cna', filters=8)
-        config['loss'] = 'mse'
+        config['body']['encoder'] = dict(base_class=None, num_stages=None, blocks=None)
```
roman-kh:
```suggestion
        config['body/encoder'] = dict(base_class=None, num_stages=None, blocks=None)
```

batchflow/models/tf/encoder_decoder.py
```diff
@@ -4,173 +4,288 @@
...
 
+            #Decoder: transition up
             x = cls.decoder(encoder_outputs, **decoder, **kwargs)
         return x
 
+
     @classmethod
     def head(cls, inputs, targets, name='head', **kwargs):
-        """ Linear convolutions with kernel 1 """
+        """ Linear convolutions with kernel size of 1. """
```
roman-kh:
```suggestion
        """ Linear convolutions. """
```

batchflow/models/tf/resnet.py
```diff
@@ -425,6 +425,13 @@ def make_encoder(cls, inputs, name='encoder', **kwargs):
         """
         num_blocks = cls.get('num_blocks', config=cls.fill_params('body', **kwargs))
 
+        if kwargs.get('filters') is None:
+            raise ValueError('Specify number of filters')
```
roman-kh: Add Raises section to docstring

batchflow/models/tf/encoder_decoder.py
```diff
@@ -326,3 +329,58 @@ def decoder(cls, inputs, name='decoder', **kwargs):
...
+            If neither `layout` nor `base` key is provided.
+        """
+        if (kwargs.get('layout') is not None) or (kwargs.get('base') is not None):
+            base_block = kwargs.get('base') or conv_block
+
+            mean = base_block(inputs, name='mean', **kwargs)
+            std = base_block(inputs, name='std', **kwargs)
+            eps = tf.random.normal(shape=tf.shape(mean))
+            return mean + eps*std
+        raise ValueError('Either `layout` or `base` key should be specified. ')
```
roman-kh: Put `raise` below `if` and move the main body outside `if` for better code clarity

batchflow/models/tf/encoder_decoder.py
```diff
@@ -326,3 +329,58 @@ def decoder(cls, inputs, name='decoder', **kwargs):
...
                         x = tf.concat((x, inputs[-i-3]), axis=axis, name='skip-concat')
         return x
+
+
+class AutoEncoder(EncoderDecoder):
+    """ Model without skip-connections between corresponding stages of encoder and decoder. """
+    @classmethod
+    def default_config(cls):
+        config = EncoderDecoder.default_config()
+        config['body/decoder'] = dict(skip=False)
```
roman-kh: Some defaults for the embedding block might be useful here, what do you think?

SergeyTsimfer: Is there any particular difference why it should be included only for autoencoders and not for encoder-decoder in general?
The reason why I did not include that in `EncoderDecoder` class is that it looks rather strange to compose encoder from massive blocks (e.g. DenseNet.block) and use small conv_block as the final encoding step.
Also, default value for number of channels is definitely ill-advised due to it's restrictive nature. For autoencoders in particular, dimensionality of compressed image is one of the most important parameters to control manually.
On the other hand, there is a possibility for not so restrictive default value of not changing number of channels, but it need to be incorporated in the `embedding` method:
```suggestion
        filters = kwargs.get('filters') or num_filters(inputs)
```

But such addition poses a new question: why should not we do the same for encoder/decoder `base` blocks?

SergeyTsimfer: Added `default_block` in the `EncoderDecoder` class. That is a shape-preserving  `conv_block` with predefined `layout`.
That block is used instead of `conv_block` as default.

batchflow/models/tf/encoder_decoder.py
```diff
@@ -4,173 +4,300 @@
...
             channels = cls.num_channels(targets)
-            x = conv_block(x, layout='c', filters=channels, kernel_size=1, **kwargs)
+            x = conv_block(x, filters=channels, **kwargs)
         return x
 
+
     @classmethod
-    def encoder(cls, inputs, base_class, name='encoder', **kwargs):
-        """ Create encoder from a base_class model
+    def default_block(cls, inputs, name='default_block', **kwargs):
```
roman-kh:
```suggestion
    def block(cls, inputs, name='default_block', **kwargs):
```
It's default by default :-)

PR #309, opened 2019-05-17
Research

AlexeyKozhevin:
- Fix bug with `grid`
- Add cross-validation
- Add parallelization for `TorchModel`
- Rename `research` methods (add `add_`)


batchflow/research/research.py
```diff
@@ -499,6 +555,16 @@ def get_copy(self):
         new_unit.variables = copy(new_unit.variables)
         return new_unit
 
+    def concat_dataset(self):
```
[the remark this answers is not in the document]
AlexeyKozhevin: I like `set`

batchflow/research/research.py
```diff
@@ -445,7 +499,7 @@ def add_function(self, function, name, execute='%1', dump=-1, returns=None,
         self._clear_result()
         self._process_iterations()
 
-    def add_pipeline(self, root_pipeline, name, branch_pipeline=None, variables=None,
+    def add_pipeline(self, root_pipeline, name, branch_pipeline=None, dataset=None, part=None, variables=None,
```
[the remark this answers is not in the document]
AlexeyKozhevin: Done

batchflow/research/job.py
```diff
@@ -6,19 +6,19 @@
 
 class Job:
     """ Contains one job. """
-    def __init__(self, executable_units, n_iters, repetition, configs, branches, name):
+    def __init__(self, executable_units, n_iters, repetition, configs, cv_splits, branches, name):
```
roman-kh:
```suggestion
    def __init__(self, executable_units, n_iters, repetition, configs, cv, branches, name):
```

batchflow/research/research.py
```diff
@@ -178,7 +184,7 @@ def load_results(self, *args, **kwargs):
         """ Load results of research as pandas.DataFrame or dict (see Results.load). """
         return Results(research=self).load(*args, **kwargs)
 
-    def _create_jobs(self, n_reps, n_iters, branches, name):
+    def _create_jobs(self, n_reps, n_iters, cv_splits, branches, name):
```
roman-kh:
```suggestion
    def _create_jobs(self, n_reps, n_iters, cv, branches, name):
```
Isn't it better or less clear??

batchflow/research/research.py
```diff
@@ -229,16 +241,27 @@ def _chunks(self, array, size):
...
+    def _cv_split(self, n_splits, shuffle):
+        has_dataset = False
+        for unit in self.executables:
+            if getattr(self.executables[unit], 'dataset', None):
+                has_dataset = True
+                self.executables[unit].dataset.cv_split(n_splits=n_splits, shuffle=shuffle)
+        if not has_dataset:
+            raise ValueError('At least one pipeline must have dataset to perform cross-validation')
+
+    def run(self, n_reps=1, n_iters=None, workers=1, branches=1, n_splits=None, shuffle=False, name=None,
```
roman-kh:
```suggestion
    def run(self, n_reps=1, n_iters=None, workers=1, branches=1, cv=None, shuffle=False, name=None,
```
??

AlexeyKozhevin: I prefer to name it `n_splits` here because it is exactly a parameter which is used in `cv_split` so it is reasonable to have the same name. Moreover, we use `cv` for folds after splitting.

batchflow/research/job.py
AlexeyKozhevin: It is not clear: `cv` can be the number of folds or indexes of folds so in the first case I use `n_splits` and in the second `cv_splits`. Maybe we should use `folds` for that parameter.

PR #312, opened 2019-05-23
Train steps

SergeyTsimfer:
Added control over

- set of variables to optimize. To use this, one can pass `scope` key to the model config. For example, `{'scope': 'body'}` would create model where only layers in `body` part are actually trained. Syntax support list of keys -- in that case, variables from union of each scope are optimized. `-` sign stands for complement of scope, e.g. `-head` means optimizing every variable not from `head` scope.

- tensors to compute loss from. To use this, one can pass parameters `predictions` and `targets` to reference respective tensors and compute loss between them, provided those tensors actually exist in model graph. For example, {'loss': {'name': 'mse', 'predictions': 'body_output', 'targets': 'some_auxiliary_target'}}.

- different `train_steps` to create multiple training procedures for one model. For example, to alternate between training only `body` layers and only `head` layers, one can use following:
```
config = {'train_steps': {'body_train': {'scope': 'body'},
                          'head_train': {'scope': 'head', 'loss': '...',
                                         'optimizer': '...', 'decay': '...'},
                                      }
}
```
Keys in `train_steps` dictionary (`body_train` and `head_train` in this case) serve as identifiers for `train` method:
```
train_pipeline = Pipeline.train_model(name, fetches, feed_dict, train_mode='body_train')
```

Also, `_check_tensor` method  is changed: now it can retrieve not only tensors that were stored via `store_to_attr`. It works if you added named `conv_block` in subclass , or added atomic `tf` named layer like `identity` or `tanh`. Full name or its unique substring  can be used as identifier. This can be helpful if you want to reference some custom layer to compute loss from (example shown above).


batchflow/models/tf/base.py
```diff
@@ -518,11 +540,50 @@ def _make_mip(self, input_name, tensor, config, depth):
...
+        if hasattr(self, name):
+            return getattr(self, name)
+
+        raise KeyError("Model %s does not have '%s' tensor" % (type(self).__name__, name))
+
+    def _make_train_steps(self, config):
+        if config.get('train_steps') is None:
+            config.update({'train_steps': {'': {key: config.get(key) for key in
+                                                ('loss', 'decay', 'optimizer', 'scope')}}})
+            total = lambda loss: tf.losses.get_total_loss()
```
roman-kh:
```suggestion
            total = lambda _: tf.losses.get_total_loss()
```

batchflow/models/tf/base.py
```diff
@@ -721,7 +806,7 @@ def _fill_output(self, output, fetches):
 
         return output
 
-    def train(self, fetches=None, feed_dict=None, use_lock=False, **kwargs):
+    def train(self, fetches=None, feed_dict=None, use_lock=False, train_mode='', **kwargs):
```
roman-kh: Allow for training all train steps at once

batchflow/models/tf/base.py
```diff
@@ -518,11 +540,50 @@ def _make_mip(self, input_name, tensor, config, depth):
...
+            return valid[0].values()[0]
+
+        valid_output = [item for item in valid if item.name.endswith('_conv_block_output')]
+        if len(valid_output) == 1:
+            return valid_output[0].values()[0]
+
+        if hasattr(self, name):
+            return getattr(self, name)
+
+        raise KeyError("Model %s does not have '%s' tensor" % (type(self).__name__, name))
```
roman-kh: Distinguish `name not found` and `many tensors match the name`

batchflow/models/tf/layers/block.py
```diff
@@ -199,6 +199,7 @@ def _conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
 
                 with tf.variable_scope('layer-%d' % i):
                     tensor = layer_fn(tensor, **args)
+    tensor = tf.identity(tensor, name='_conv_block_output')
```
roman-kh:
```suggestion
    tensor = tf.identity(tensor, name='_output')
```

PR #315, opened 2019-05-31
DeepGalerkin: inverse problem

SergeyTsimfer:
Currently, `DeepGalerkin` model infers PDE from one key `form`, which contains `d0`, `d1` and `d2` keys to communicate exact coefficients of the equation. This PR proposes to add two more forms, namely:

- `form_noise` in order to slightly distort every coefficient of the network randomly on each train iteration. That is done in order to make solution more robust. This key has the same structure, as `form`, and every value in it defines scale of applied noise. For example,
```
{'form_noise': {'d0': 0.2}}
```
applies Normal(0, 0.2) noise to the coefficient before `u(x)`. Common parameters (e.g. distribution to draw points from) for all noise additions can be configured via `pde/noise_block` key.

- `form_add` allows to add trainable networks to each coefficient of the PDE. Again, it has the same structure as `form` key, but every value just determines whether to add trainable network to the coefficient or not. Common parameters, like layout of added network (or any other `conv_block` parameter), can be configured via `pde/addendum_block` key.

Both these forms allow to directly apply noise/network to right-hand side by passing `rhs` key to them.

Also, added `additional_lhs` and `additional_rhs` keys. The first one allows to store additional expressions of the network in desired place, for example:
```
'additional_lhs': {'my_expr': {'form': {..}}}
```
allows to store any expression, supported by `form`/`form_noise`/`form_add` syntax. Name `my_expr` is latter accessible to be `fetch`'ed by `predict` method or can be used as prediction for loss computation(more on that later).

Key `additional_rhs` is used to store `tf.callable`'s that can be accessed later (same as previous key). For example,
```
'additional_rhs': {'cond_1': lambda t: tf.sin(K*t)*tf.exp(t)+2}
```

To compute loss between two additional values, we can use:
```
'loss': {'name': 'mse', 'predictions': 'my_expr', 'targets': 'cond_1'}
```

With the usage of `train_steps` key we can easily train network to account for different conditions, different scopes (note that each additional network created by `form_add` lies in the predefined  `addendums` scope) and so on. Complete example can be seen [here](https://github.com/analysiscenter/batchflow/blob/24b4af0b9040719f7aa0ce4f2f3f871ab3af5630/examples/experiments/DG_inverse.ipynb)


batchflow/models/tf/deep_galerkin.py
```diff
@@ -433,9 +528,95 @@ def _callable(shifted_time):
...
+class DGSolver():
+    """ Wrapper around `DeepGalerkin` to surpass BatchFlow's syntax sugar.
+
+    Parameters
+    ----------
+    model_config : dict
+        Configuration of model. Supports all of the options from `DeepGalerkin`.
+    """
+
+    def __init__(self, model_config, dg_class=DeepGalerkin):
```
roman-kh:
```suggestion
    def __init__(self, model_class=DeepGalerkin, config=None):
```
To comply with Pipeline API

batchflow/models/tf/deep_galerkin.py
```diff
@@ -433,9 +528,95 @@ def _callable(shifted_time):
...
+        self.model = dg_class(model_config)
```
roman-kh:
```suggestion
        self.model = model_class(config=config)
```

batchflow/models/tf/deep_galerkin.py
```diff
@@ -433,9 +528,95 @@ def _callable(shifted_time):
...
     def predict(self, fetches=None, feed_dict=None, **kwargs):
         """ Get network-approximation of PDE-solution on a set of points. Overloads :meth:`.TFModel.predict` :
         `solution`-tensor is now considered to be the main model-output.
         """
         fetches = 'solution' if fetches is None else fetches
         return super().predict(fetches, feed_dict, **kwargs)
+
+
+
+class DGSolver():
```
roman-kh:
```suggestion
class DGSolver:
```

roman-kh: Perhaps, we could make it compatible with BaseModel API to use this class in Pipelines as well

batchflow/models/tf/deep_galerkin.py
```diff
@@ -433,9 +528,95 @@ def _callable(shifted_time):
...
+    ----------
+    model_config : dict
+        Configuration of model. Supports all of the options from `DeepGalerkin`.
+    """
+
+    def __init__(self, model_config, dg_class=DeepGalerkin):
+        self.model = dg_class(model_config)
+
+
+    def fit(self, sampler, batch_size, n_iters, train_mode='', fetches=None, bar=False):
```
roman-kh:
```suggestion
    def fit(self, batch_size, sampler, n_iters, *args, train_mode='', fetches=None, bar=False):
```

SergeyTsimfer:
Contents of this PR are fully incorporated into #330.

PR #321, opened 2019-06-17
Fix case of torch model single input

nikita-klsh:
If len(args) == 1, then, for example:

*inputs, targets = tuple([1])
print(inputs, targets)
<< [], 1

I believe this piece of code is not supposed to perform that way.


roman-kh:
I've implemented this with a keyword arg `targets` in 8eb6087.
Do you think having a positional arg for `targets` is also needed?

PR #328, opened 2019-07-02
DeepLab v3+

SergeyTsimfer:
This PR proposes following additions:

1. Change `ASPP` module in order to make it compatible with dynamic shapes. Note that building model with dynamic `ASPP` is slower, but there is no difference at train time.
2. Add [`Xception`](https://arxiv.org/pdf/1610.02357.pdf) model architecture: authors make case for heavy usage of `depthwise_conv`-`batchnorm`-`pointwise_conv` blocks. Note that modifications, pointed out in [this](http://presentations.cocodataset.org/COCO17-Detect-MSRA.pdf) paper, are applied, as well as modifications from this [implementation.](https://github.com/tensorflow/models/blob/58a3de6c68639c3eac95f7c84089f320d32a85bb/research/deeplab/core/xception.py)
3. Add [`DeepLab`](https://arxiv.org/pdf/1802.02611.pdf) model architecture: this one uses `Xception` as its backbone, followed by `ASPP` module and simple yet effective decoder. Note that only v3+ version is implemented: previous ones are either too heavy (due to CRF usage) or less effective (due to lack of decoder). Currently `Xception` blocks are used by default; it is easy to swap them with, for example, `MobileNet v2` inverted residual blocks, and I am in search for better syntax for doing so.

### DIFF
--- a/batchflow/models/tf/deeplab.py
+++ b/batchflow/models/tf/deeplab.py
@@ -80,10 +80,10 @@ class DeepLabX(DeepLab):
     def default_config(cls):
         config = super().default_config()
         config['body/encoder/base'] = Xception
-        config['body/encoder/entry'] = dict(num_stages=None, filters=None, strides=2, combine_type='conv')
-        config['body/encoder/middle'] = dict(num_stages=None, filters=None, strides=1, combine_type='sum')
+        config['body/encoder/entry'] = dict(num_stages=None, filters=None, strides=2, combine_op='softsum')
+        config['body/encoder/middle'] = dict(num_stages=None, filters=None, strides=1, combine_op='sum')
         config['body/encoder/exit'] = dict(num_stages=None, filters=None, strides=1,
-                                           depth_activation=True, combine_type='conv')
+                                           depth_activation=True, combine_op='softsum')
         config['body/encoder/downsample'] = dict(layout=None)
         return config
 
--- a/batchflow/models/tf/deeplab.py
+++ b/batchflow/models/tf/deeplab.py
@@ -43,7 +43,6 @@ class DeepLab(EncoderDecoder):
         config['body/entry/downsample'] = dict(layout=None)
         config['body/entry/blocks'] = dict(base=Xception.block, filters=None, strides=2, combine_type='conv')
 
-
         config['body/middle'] = dict(base=Xception.block, num_stages=None,
                                      filters=None, strides=1, combine_type='sum')
 
@@ -115,7 +114,7 @@ class DeepLab8(DeepLab):
     @classmethod
     def default_config(cls):
         config = super().default_config()
-        config['initial_block']: dict(layout='cnacna', filters=[32], kernel_size=3, strides=1)
+        config['initial_block'] = dict(layout='cnacna', filters=32, kernel_size=3, strides=1)
 
         config['body/entry/num_stages'] = 4
         config['body/entry/blocks/filters'] = [[64]*3,
@@ -144,7 +143,7 @@ class DeepLab16(DeepLab):
     @classmethod
     def default_config(cls):
         config = super().default_config()
-        config['initial_block']: dict(layout='cnacna', filters=[32], kernel_size=3, strides=1)
+        config['initial_block'] = dict(layout='cnacna', filters=32, kernel_size=3, strides=1)
 
         config['body/entry/num_stages'] = 4
         config['body/entry/blocks/filters'] = [[64]*3,
@@ -173,7 +172,7 @@ class DeepLabS(DeepLab):
     @classmethod
     def default_config(cls):
         config = super().default_config()
-        config['initial_block']: dict(layout='cna', filters=[32], kernel_size=3, strides=1)
+        config['initial_block'] = dict(layout='cna', filters=32, kernel_size=3, strides=1)
 
         config['body/entry/num_stages'] = 4
         config['body/entry/blocks/filters'] = [[6]*3,
--- a/batchflow/models/tf/deeplab.py
+++ b/batchflow/models/tf/deeplab.py
@@ -11,9 +11,8 @@ from .layers import aspp
 
 class DeepLab(EncoderDecoder):
     """ DeepLab v3+ model archtecture.
-    By default, input is encoded with Xception backbone (entry, backbone_middle and backbone_exit flows),
-    then worked via Atrous Spatial Pyramid Pooling layer, and, finally, restored to original
-    spatial size with simple yet effective decoder module.
+    By default, input is compressed with encoder, then worked via Atrous Spatial Pyramid Pooling layer,
+    and, finally, restored to original spatial size with simple yet effective decoder module.
 
     Parameters
     ----------
@@ -21,34 +20,23 @@ class DeepLab(EncoderDecoder):
         Dictionary with 'images' (see :meth:`~.TFModel._make_inputs`)
 
     body : dict
-        entry : dict
+        encoder : dict
         Dictionary with parameters for entry encoding: downsampling of the inputs.
         See :meth:`~.EncoderDecoder.encoder` arguments.
 
-        backbone_middle : dict
-        Dictionary with parameters for backbone_middle encoding: thorough processing.
-        See :meth:`~.EncoderDecoder.embedding` arguments.
-
-        backbone_exit : dict
-        Dictionary with parameters for backbone_exit encoding: final increase in number of features.
-        See :meth:`~.EncoderDecoder.embedding` arguments.
-
         embedding : dict
         Same as :meth:`~.EncoderDecoder.embedding`. Default is to use ASPP.
+
+        decoder : dict
+        Dictionary with parameters for decoding of compressed representation.
+        See :meth:`~.EncoderDecoder.decoder` arguments. Default is to use bilinear upsampling.
     """
     @classmethod
     def default_config(cls):
         config = super().default_config()
         config['body/encoder'] = dict(base=None, num_stages=None)
-        config['body/encoder/downsample'] = dict(layout=None)
-        config['body/encoder/blocks'] = dict(base=Xception.block, filters=None, strides=2, combine_type='conv')
 
-        config['body/embeddings/common'] = dict(base=Xception.block)
-        config['body/backbone_middle'] = dict(num_stages=None, filters=None, strides=1, combine_type='sum')
-        config['body/backbone_exit'] = dict(num_stages=None, filters=None, strides=1,
-                                            depth_activation=True, combine_type='conv')
         config['body/embedding'] = dict(base=aspp)
-        config['body/embeddings/order'] = ['backbone_middle', 'backbone_exit', 'embedding']
 
         config['body/decoder'] = dict(skip=True, num_stages=None, factor=None)
         config['body/decoder/upsample'] = dict(layout='b')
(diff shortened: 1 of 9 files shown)


batchflow/models/tf/deeplab.py
```diff
@@ -0,0 +1,198 @@
...
+"""
+import tensorflow as tf
+
+from .layers import aspp
+from . import EncoderDecoder, Xception
+
+
+
+class DeepLab(EncoderDecoder):
+    """ DeepLab model archtecture.
```
roman-kh: Worth mentioning here that it's a V3+ arch.

batchflow/models/tf/layers/pyramid.py
```diff
@@ -6,7 +6,7 @@
 
 
 def pyramid_pooling(inputs, layout='cna', filters=None, kernel_size=1, pool_op='mean', pyramid=(0, 1, 2, 3, 6),
-                    name='psp', **kwargs):
+                    reshape_to_vector=False, name='psp', **kwargs):
```
roman-kh:
```suggestion
                    flatten=False, name='psp', **kwargs):
```

batchflow/models/tf/layers/pyramid.py
```diff
@@ -115,7 +176,7 @@ def aspp(inputs, layout='cna', filters=None, kernel_size=3, rates=(6, 12, 18), i
 
         for level in rates:
             x = conv_block(inputs, layout, filters=filters, kernel_size=kernel_size, dilation_rate=level,
-                           name='conv-%d' % level, **kwargs)
+                           name='atrous_conv-%d' % level, **kwargs)
```
roman-kh: Why is `conv` not OK?

batchflow/models/tf/xception.py
```diff
@@ -0,0 +1,219 @@
...
+    @classmethod
+    def slice_dict(cls, dictionary, idx):
+        return {key: value[idx] for key, value in dictionary.items()
+                if isinstance(value, list)}
+
+
+    @classmethod
+    def body(cls, inputs, name='body', **kwargs):
+        """ Entry, middle and exit flows consequently. """
+        kwargs = cls.fill_params('body', **kwargs)
```
roman-kh: Let's rename `fill_params` to reconcile with TorchModel

batchflow/models/tf/deeplab.py
```diff
@@ -0,0 +1,198 @@
...
+    Input is encoded through Xception backbone (entry, middle and exit flows),
```
roman-kh:
```suggestion
    By default, input is encoded through Xception backbone (entry, middle and exit flows),
```

batchflow/models/tf/deeplab.py
```diff
@@ -0,0 +1,198 @@
...
+    then worked via Atrous Spatial Pyramid Pooling layer, and, finally, restored to original
+    spatial size with simple yet effective decoder module.
+
+    Parameters
+    ----------
+    inputs : dict
+        Dictionary with 'images' (see :meth:`~.TFModel._make_inputs`)
+
+    body : dict
+        entry : dict
```
roman-kh: In general, `deeplab` has nothing to do with entry, middle, exit architecture. These are all Xception block parameters, but deeplab can be built upon ResNet backbone.

batchflow/models/tf/layers/pyramid.py
SergeyTsimfer: Was using it to differentiate between ASPP layer convolutions and others during debugging. Removed

batchflow/models/tf/xception.py
```diff
@@ -0,0 +1,199 @@
...
+
+            x = conv_block(x, layout, filters=filters, kernel_size=1, **kwargs)
+        return x
+
+
+    @classmethod
+    def _depthwise_conv(cls, inputs, depth_multiplier=1, name='depthwise_conv', **kwargs):
+        data_format = kwargs.get('data_format')
+        filters = cls.num_channels(inputs, data_format)
+        depthwise_conv = tf.keras.layers.SeparableConv2D(filters=filters,
```
roman-kh: This only works with 2d data, but we definitely need Nd solution. There is separable conv in convblock. Does it fit the purpose?

batchflow/models/tf/deeplab.py
SergeyTsimfer: Fairly speaking, current implementation of `DeepLab` class has very little to do with actual DeepLab architecture: this class is merely `EncoderDecoder` with multiple `embedding` blocks, default config values and some custom names. My plan is to:
1. Add `make_encoder` method to `Xception` class. That would allow to replicate one of the architectures from the article via `EncoderDecoder` only with config options: `Xception` as encoder and `ASPP` module as embedding.
2. Rename `entry`/`middle`/`exit` blocks in `DeepLab`, so it stays highly customizable, but without references to `Xception`

batchflow/models/tf/xception.py
SergeyTsimfer: The reason I need separate function for `depthwise` conv is to insert other layers (namely, batch-norm and activation) between depthwise and pointwise convolutions. Current implementation of separable convolution does not allow to do so;
Furthermore, `keras` implementation is 2 times faster than the previous one at build time, and 35% faster at train time.
I can make `depthwise_conv` function inside conv.py, which would look pretty similar to `conv` function: get number of dimensions and choose appropriate layer to apply

SergeyTsimfer:
Previously, `_separate_conv` function inside `conv.py` was doing, essentially, two things: depthwise convolution via loop on channels, and pointwise convolution via `conv` layer.
I've made a separate function for the first part: it allows to fallback on better-optimized `tf.keras.layers.DepthwiseConv2D` if needed and to use this function not as a part of `separable_convolution`.

batchflow/models/tf/deeplab.py
SergeyTsimfer: So I decided to make `EncoderDecoder` class capable of handling multiple `embedding`'s. That definetely belongs to that class, not to `DeepLab`. It allowed to make `DeepLab` class by only changing `default_config`.
Commit is on the way..

SergeyTsimfer:
Moved multiple embeddings functionality to `EncoderDecoder` class. Now, there are two more additional config-options: `body/embeddings/common` and `body/embeddings/order`. `DeepLab` architecture is now just a config

batchflow/models/tf/deeplab.py
roman-kh: Deeplab config should not have entry, middle and exit options as these belong to Xception backbone, but not Deeplab itself. Deeplab can be built upon ResNet backbone (or any other architecture)

batchflow/models/tf/encoder_decoder.py
```diff
@@ -111,17 +112,21 @@ class EncoderDecoder(TFModel):
...
     def default_config(cls):
         config = TFModel.default_config()
+
         config['body/encoder'] = dict(base=None, num_stages=None)
         config['body/encoder/downsample'] = dict(layout='p', pool_size=2, pool_strides=2)
         config['body/encoder/blocks'] = dict(base=cls.block)
-        config['body/embedding'] = dict(base=cls.block)
+
+        config['body/embedding'] = dict()
+        config['body/embeddings/common'] = dict(base=cls.block)
```
roman-kh: Having both `embedding` and `embeddings` is very confusing. Let's allow `embedding` to be a list of dicts when configuring multiple embeddings

batchflow/models/tf/xception.py
```diff
@@ -0,0 +1,214 @@
...
+        kwargs = cls.fill_params('body', **kwargs)
+        entry = kwargs.pop('entry')
+        middle = kwargs.pop('middle')
+        exit = kwargs.pop('exit')
+
+        with tf.variable_scope(name):
+            x = inputs
+
+            # Entry flow: downsample the inputs
+            with tf.variable_scope('entry_flow'):
```
roman-kh: Why not just `entry`?

batchflow/models/tf/xception.py
```diff
@@ -0,0 +1,214 @@
...
+                entry_stages = entry.pop('num_stages', 0)
+                for i in range(entry_stages):
+                    with tf.variable_scope('group-'+str(i)):
+                        args = {**kwargs, **entry, **unpack_args(entry, i, entry_stages)}
+                        x = cls.block(x, name='block-'+str(i), **args)
+                        x = tf.identity(x, name='output')
+
+            # Middle flow: thorough processing
+            with tf.variable_scope('middle_flow'):
```
roman-kh: Why not just `middle`?

batchflow/models/tf/xception.py
```diff
@@ -0,0 +1,214 @@
...
+                middle_stages = middle.pop('num_stages', 0)
+                for i in range(middle_stages):
+                    args = {**kwargs, **middle, **unpack_args(middle, i, middle_stages)}
+                    x = cls.block(x, name='block-'+str(i), **args)
+
+            # Exit flow: final increase in number of feature maps
+            with tf.variable_scope('exit_flow'):
```
roman-kh: Why not just `exit`?

batchflow/models/tf/deeplab.py
```diff
@@ -0,0 +1,186 @@
...
+    @classmethod
+    def default_config(cls):
+        config = super().default_config()
+        config['body/encoder'] = dict(base=None, num_stages=None)
+
+        config['body/embedding'] = dict(base=aspp)
+
+        config['body/decoder'] = dict(skip=True, num_stages=None, factor=None)
+        config['body/decoder/upsample'] = dict(layout='b')
+        config['body/decoder/blocks'] = dict(base=cls.block)
```
roman-kh: This is not needed, as it repeats the parent class config

batchflow/models/tf/encoder_decoder.py
```diff
@@ -251,9 +259,16 @@ def embedding(cls, inputs, name='embedding', **kwargs):
         -------
         tf.Tensor
         """
+        steps = kwargs.get('num_stages', 1)
         base_block = kwargs.get('base')
```
roman-kh:
```suggestion
        base_block = kwargs.get('base', conv_block)
```

batchflow/models/tf/encoder_decoder.py
```diff
@@ -368,9 +387,15 @@ def embedding(cls, inputs, name='embedding', **kwargs):
...
+        x = inputs
 
-        mean = base_block(inputs, name='mean', **kwargs)
-        std = base_block(inputs, name='std', **kwargs)
-        eps = tf.random.normal(shape=tf.shape(mean))
-        return mean + eps*std
+        with tf.variable_scope(name):
+            for i in range(steps):
```
roman-kh: Why do you need steps here?

batchflow/models/tf/encoder_decoder.py
```diff
@@ -251,9 +259,16 @@ def embedding(cls, inputs, name='embedding', **kwargs):
         -------
         tf.Tensor
         """
+        steps = kwargs.get('num_stages', 1)
```
roman-kh: Having `step` and a list of embeddings seems like a total overkill. Let's choose only one of those or even reject both in favor of method overloading

batchflow/models/tf/xception.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        steps = 1 + cls.get('entry/num_stages', config=cls.fill_params('body', **kwargs))
+
+        with tf.variable_scope(name):
+            x = cls.body(inputs, name='body', **kwargs)
+
+            scope = tf.get_default_graph().get_name_scope()
+            encoder_tensors = [inputs]
+            for i in range(steps):
+                tensor_name = scope + '/body/entry/group-'+str(i) + '/output:0'
```
roman-kh: Let's create a default TFModel's `make_encoder` which build body and the gets tensors for a given regexp

batchflow/models/tf/encoder_decoder.py
SergeyTsimfer: Looks like it is quite easy to replicate `num_stages` parameters via list of embeddings, so I've decided to remove it.

PR #333, opened 2019-07-11
Mobilenetv3

dmylzenova:


batchflow/models/tf/base.py
```diff
@@ -1940,9 +1941,13 @@ def se_block(cls, inputs, ratio, name='se', **kwargs):
         with tf.variable_scope(name):
             data_format = kwargs.get('data_format')
             in_filters = cls.num_channels(inputs, data_format)
+            if isinstance(kwargs.get('activation'), list) and len(kwargs.get('activation')) == 2:
+                activation = kwargs.pop('activation')
+            else:
+                activation = [tf.nn.relu, tf.nn.sigmoid]
```
roman-kh: This ignores the passed activation while it should not

batchflow/models/tf/mobilenet.py
```diff
@@ -4,6 +4,9 @@
 Sandler M. et al. "`Inverted Residuals and Linear Bottlenecks:
 Mobile Networks for Classification, Detection and Segmentation
 <https://arxiv.org/abs/1801.04381>`_"
+
+Howard A. et al. "`Searching for MobileNetV3
+<https://arxiv.org/pdf/1905.02244.pdf>`_"
```
roman-kh: Never make links to pdf, only to asbtracts

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,229 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+
+        return config
+
+    def build_config(self, names=None):
+        config = super().build_config(names)
+        config['head']['filters'][-1] = self.num_classes('targets')
+        return config
+
+    @classmethod
+    def h_swish(cls, inputs):
```
roman-kh: Move it to `models.tf.nn`

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,229 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+            input tensor
+
+        Returns
+        -------
+        tf.Tensor
+        """
+        return inputs * tf.nn.relu6(inputs + 3) / 6.0
+
+    @classmethod
+    def h_sigmoid(cls, inputs):
```
roman-kh: Move it to `models.tf.nn`

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,229 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+                for k in range(repeats):
+                    if k > 0:
+                        block['strides'] = 1
+                    residual = block.pop('residual', False) or k > 0
+                    x = cls.block(x, **block, residual=residual, name='block-%d' % i, **kwargs)
+                    i += 1
+        return x
+
+    @classmethod
+    def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=4, width_factor=1, h_swish=False,
```
roman-kh: You don't need `h_swish` here, activation seems more than enough

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,229 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+        name : str
+            scope name
+
+        Returns
+        -------
+        tf.Tensor
+        """
+
+        if h_swish:
+            kwargs['activation'] = cls.h_swish # will not work in current se_block
```
roman-kh: If it does not work, there is no reason to write like this

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,229 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+
+        Returns
+        -------
+        tf.Tensor
+        """
+        kwargs = cls.fill_params('head', **kwargs)
+        layout = kwargs.pop('layout')
+        filters = kwargs.pop('filters')
+        if kwargs.get('se_block'):
+            x = conv_block(inputs, layout=layout[:3], filters=filters[0], name=name, **kwargs)
```
roman-kh: This makes head config very unclear and confusing. Consider splitting it.

batchflow/models/tf/mobilenet.py
dmylzenova: moved to models.tf.nn

batchflow/models/tf/base.py
dmylzenova: it ignores the passed activation only in case it wasn't supposed to be used in the se_block. Two lines above I check whether activation is a list of two elements since se_block needs two.
By default, se_block will get activation from common config, therefore, activations were hardcoded before. Mobilenetv3 needs to change tf.nn.sigmoid to piece-wise linear analog.

batchflow/models/tf/mobilenet.py
dmylzenova: Both splitting layout in the head config to several and splitting head moving se_block to the body don't seem good to me. Please suggest a variant that seems less confusing to you.

batchflow/models/tf/base.py
dmylzenova: The passed activation will be ignored only in case it wasn't supposed to be used in se_block. Two lines above I check whether activation is a list of two elements needed for se_block. 
Se_block can get an activation function shared with other layers, for example, in the current resnet use case.

roman-kh: If the passed activation is not a list of two, it will be silently ignored at all. Which is not a good idea. You could have a default list of `[tf.nn.relu, tf.nn.sigmoid]` and replace the first item with the passed activation. Or you could throw an exception. Or a warning.

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,194 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+            residual : bool
+
+        width_factor : float
+            multiplier for the number of channels (default=1)
+    """
+    @classmethod
+    def default_config(cls):
+        config = TFModel.default_config()
+        config['initial_block'].update(dict(layout='cna', filters=16, kernel_size=3, strides=2, activation=h_swish))
+        config['body'].update(dict(width_factor=1, layout=deepcopy(_V3_LARGE_DEFAULT_BODY)))
```
roman-kh: Why do you need deepcopy here?

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,194 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+    dict(repeats=3, filters=96, expansion_factor=6, strides=2, kernel_size=5, activation=h_swish, se_block=True),
+]
+
+class MobileNet_v3_small(MobileNet_v3):
+    """ MobileNet version 3 small architecture """
+    @classmethod
+    def default_config(cls):
+        config = MobileNet_v3.default_config()
+        config['initial_block'].update(dict(layout='cna', filters=16, kernel_size=3, strides=2, activation=h_swish))
+        config['body'].update(dict(width_factor=1, layout=deepcopy(_V3_SMALL_DEFAULT_BODY)))
```
roman-kh: Why do you need deepcopy here?

batchflow/models/tf/mobilenet.py
roman-kh: You could make `se_block` a dict with all the needed params: filters, layout, activations, etc

batchflow/models/tf/mobilenet.py
```diff
@@ -224,3 +227,197 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+
+        num_filters = int(cls.num_channels(inputs, kwargs.get('data_format')) * expansion_factor * width_factor)
+        x = conv_block(inputs, 'cna', num_filters, 1, name='%s-exp' % name,
+                       strides=1, **kwargs)
+        x = depthwise_conv(x, kernel_size=kernel_size, strides=strides,
+                           padding='same', data_format=kwargs['data_format'], name='depthwise')
+        conv_block(x, 'na', name='%s-exp-a' % name, **kwargs)
+
+        if se_block:
+            x = cls.se_block(x, int(num_filters * 0.25), name='%s-se' % name, data_format=kwargs['data_format'],
```
roman-kh:
```suggestion
            x = cls.se_block(x, num_filters // 0.25, name='%s-se' % name, data_format=kwargs['data_format'],
```

batchflow/models/tf/base.py
dmylzenova: added warning

batchflow/models/tf/mobilenet.py
dmylzenova: done

batchflow/models/tf/mobilenet.py
dmylzenova: because layout is modified in the body.

batchflow/models/tf/mobilenet.py
dmylzenova: because layout is modified in the body.

batchflow/models/tf/mobilenet.py
dmylzenova: done

batchflow/models/tf/mobilenet.py
dmylzenova: done

batchflow/models/tf/base.py
roman-kh: Why not `activation = [activation, tf.nn.sigmoid]`?

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +222,203 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
         tf.Tensor
         """
         num_filters = int(cls.num_channels(inputs, kwargs.get('data_format')) * expansion_factor * width_factor)
-        conv_filters = [num_filters, num_filters, filters]
-        x = conv_block(inputs, 'cna Cna cn', conv_filters, [1, 3, 1], name=name, strides=[1, strides, 1], **kwargs)
+
+        x = conv_block(inputs, 'cna', num_filters, kernel_size=1, name='%s-exp' % name, strides=1, **kwargs)
+        x = depthwise_conv(x, kernel_size=3, strides=strides, padding='same',
+                           data_format=kwargs['data_format'], name='%s-depthwise' % name)
+        x = conv_block(x, 'nacn', filters, kernel_size=1, strides=1, name='%s-down' % name, **kwargs)
```
roman-kh: What's the point in splitting 1 call of `conv_block` into 3 calls?

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +222,203 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+
+        num_filters = int(cls.num_channels(inputs, kwargs.get('data_format')) * expansion_factor * width_factor)
+        x = conv_block(inputs, 'cna', num_filters, 1, name='%s-exp' % name,
+                       strides=1, **kwargs)
+        x = depthwise_conv(x, kernel_size=kernel_size, strides=strides,
+                           padding='same', data_format=kwargs['data_format'], name='%s-depthwise' % name)
+        x = conv_block(x, 'na', name='%s-na' % name, **kwargs)
+
+        if se_block:
+            x = cls.se_block(x, num_filters // 4, name='%s-se' % name, data_format=kwargs['data_format'],
```
roman-kh: Sometimes you use `kwargs['data_format']`, but otherwise `kwargs.get('data_format')`. Be consistent

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +222,203 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+        config['body'].update(dict(width_factor=1, layout=deepcopy(_V3_LARGE_DEFAULT_BODY)))
+        config['head'].update(dict(layout='cnavcacV', filters=[960, 1280, 2], pool_size=7, kernel_size=1,
+                                   activation=h_swish))
+        config['loss'] = 'ce'
+
+        return config
+
+    def build_config(self, names=None):
+        config = super().build_config(names)
+        config['head']['filters'][-1] = self.num_classes('targets')
```
roman-kh:
```suggestion
        config['head/filters'][-1] = self.num_classes('targets')
```

roman-kh: This will fail if head has different layout (e.g. 'cnaV')

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +222,203 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+            number of output filters
+        residual : bool
+            whether to make a residual connection
+        strides : int
+            stride for 3x3 convolution
+        expansion_factor : int
+            multiplier for the number of filters in internal convolutions
+        width_factor : float
+            multiplier for the number of filters
+        h_swish : bool
```
roman-kh: Nope

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +222,203 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+        config = MobileNet_v3.default_config()
+        config['initial_block'].update(dict(layout='cna', filters=16, kernel_size=3, strides=2, activation=h_swish))
+        config['body'].update(dict(width_factor=1, layout=deepcopy(_V3_SMALL_DEFAULT_BODY)))
+        config['head'].update(dict(layout='cnavcacV', filters=[576, 1280, 2], pool_size=7,
+                                   kernel_size=1, activation=h_swish, se_block=True,
+                                   se_kwargs=dict(activation=[h_swish, h_sigmoid], ratio=144)))
+        return config
+
+    @classmethod
+    def head(cls, inputs, se_kwargs=None, name='head', **kwargs):
```
roman-kh:
```suggestion
    def head(cls, inputs, se_block=None, name='head', **kwargs):
```

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +222,203 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+        Returns
+        -------
+        tf.Tensor
+        """
+        kwargs = cls.fill_params('head', **kwargs)
+        layout = kwargs.pop('layout')
+        filters = kwargs.pop('filters')
+        if kwargs.get('se_block'):
+            x = conv_block(inputs, layout=layout[:3], filters=filters[0], name='%s-conv1' % name, **kwargs)
+            x = cls.se_block(x, **se_kwargs, data_format=kwargs['data_format'], name='%s-se' % name)
```
roman-kh: You don't need many args for the same need. `se_block` is enough for everything.

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +222,203 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+
+    body : dict
+        layout : list of dict
+            a sequence of block parameters:
+            repeats : int
+            filters : int
+            expansion_factor : int
+            strides : int
+            kernel_size : int
+            h_swish : bool
```
roman-kh: Nope

batchflow/models/tf/mobilenet.py
dmylzenova: because C performs depthwise+pointwise when only depthwise is needed. We don't have a letter for depthwise_conv only.

batchflow/models/tf/mobilenet.py
dmylzenova: ok

batchflow/models/tf/mobilenet.py
dmylzenova: why? this line changes defined in the default config list of filters. No conflicts If a layout has only one convolution - its num_filters will be changed.

batchflow/models/tf/base.py
dmylzenova: Because it is not enough for the current case and may show bad behavior if it is passed here as default for other blocks.

batchflow/models/tf/mobilenet.py
roman-kh: `head/filters` can be just `int`, not list

batchflow/models/tf/mobilenet.py
roman-kh: That's not true. Our layers perform pw-conv only if needed. And there is tf layer for dw-only conv. So we can easily add a letter for that.

batchflow/models/tf/mobilenet.py
```diff
@@ -179,48 +217,123 @@ def body(cls, inputs, name='body', **kwargs):
...
             whether to make a residual connection
         strides : int
             stride for 3x3 convolution
         expansion_factor : int
             multiplier for the number of filters in internal convolutions
         width_factor : float
             multiplier for the number of filters
+        activation : callable, optional
+            If not specified tf.nn.relu is used.
+        se_block : bool
```
roman-kh: Make it a dict to allow for custom parameters

batchflow/models/tf/mobilenet.py
```diff
@@ -134,22 +163,31 @@ class MobileNet_v2(TFModel):
...
         width_factor : float
             multiplier for the number of channels (default=1)
     """
     @classmethod
     def default_config(cls):
         config = TFModel.default_config()
-        config['common'].update(dict(activation=tf.nn.relu6))
         config['initial_block'].update(dict(layout='cna', filters=32, kernel_size=3, strides=2))
-        config['body'].update(dict(width_factor=1, layout=deepcopy(_V2_DEFAULT_BODY)))
+        config['body'].update(dict(width_factor=1, layout=deepcopy(_V2_DEFAULT_BODY), activation=tf.nn.relu6))
```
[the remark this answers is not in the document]
roman-kh: Moreover, you don't need this since you pass body kwargs to block.

batchflow/models/tf/layers/block.py
```diff
@@ -39,6 +40,7 @@
     'c': 'conv',
     't': 'transposed_conv',
     'C': 'separable_conv',
+    'w': 'depthwise_conv',
```
roman-kh: Let's make it `W` to align with separable `C` and `T`

batchflow/models/tf/mobilenet.py
```diff
@@ -134,31 +165,44 @@ class MobileNet_v2(TFModel):
...
     """
     @classmethod
     def default_config(cls):
         config = TFModel.default_config()
-        config['common'].update(dict(activation=tf.nn.relu6))
         config['initial_block'].update(dict(layout='cna', filters=32, kernel_size=3, strides=2))
         config['body'].update(dict(width_factor=1, layout=deepcopy(_V2_DEFAULT_BODY)))
         config['head'].update(dict(layout='cnacnV', filters=[1280, 2], kernel_size=1))
-
+        config['common'].update(dict(activation=tf.nn.relu6))
```
roman-kh: This will make all activation `relu6` (i.e. in intitial block and head).
If you wanted to have `relu6` in body only, pass it to body

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +232,112 @@ def body(cls, inputs, name='body', **kwargs):
...
             whether to make a residual connection
         strides : int
             stride for 3x3 convolution
         expansion_factor : int
             multiplier for the number of filters in internal convolutions
         width_factor : float
             multiplier for the number of filters
+        activation : callable, optional
+            If not specified tf.nn.relu is used.
+        se_block : bool
```
roman-kh:
```suggestion
        se_block : bool or dict
```

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +232,112 @@ def body(cls, inputs, name='body', **kwargs):
...
         num_filters = int(cls.num_channels(inputs, kwargs.get('data_format')) * expansion_factor * width_factor)
-        conv_filters = [num_filters, num_filters, filters]
-        x = conv_block(inputs, 'cna Cna cn', conv_filters, [1, 3, 1], name=name, strides=[1, strides, 1], **kwargs)
+        x = conv_block(inputs, 'cnawna', [num_filters, num_filters], [1, kernel_size], strides=[1, strides],
+                       name='%s-exp' % name, **kwargs)
+
+        if se_block:
+            if not isinstance(se_block, dict):
+                se_block = dict(activation=[kwargs.get('activation', tf.nn.relu), h_sigmoid], ratio=num_filters // 4)
+            x = cls.se_block(x, **se_block, name='%s-se' % name, data_format=kwargs.get('data_format'))
```
roman-kh: pass all kwargs, not only data format

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +232,112 @@ def body(cls, inputs, name='body', **kwargs):
...
+            scope name
+
+        Returns
+        -------
+        tf.Tensor
+        """
+        kwargs = cls.fill_params('head', **kwargs)
+        layout = kwargs.pop('layout')
+        filters = kwargs.pop('filters')
+        if kwargs.get('se_block'):
```
roman-kh: It will always be None

roman-kh: And it has never been documented anywhere that head could contain se block

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +232,112 @@ def body(cls, inputs, name='body', **kwargs):
                 for k in range(repeats):
                     if k > 0:
                         block['strides'] = 1
-                    x = cls.block(x, **block, residual=k > 0, name='block-%d' % i, **kwargs)
+                    residual = block.pop('residual', False) or k > 0
+                    x = cls.block(x, **block, residual=residual, name='block-%d' % i, **kwargs)
```
roman-kh: block and kwargs can contain same parameters

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +232,112 @@ def body(cls, inputs, name='body', **kwargs):
...
+        """
+        kwargs = cls.fill_params('head', **kwargs)
+        layout = kwargs.pop('layout')
+        filters = kwargs.pop('filters')
+        if kwargs.get('se_block'):
+            x = conv_block(inputs, layout=layout[:3], filters=filters[0], name='%s-conv1' % name, **kwargs)
+            x = cls.se_block(x, **se_block, data_format=kwargs.get('data_format'), name='%s-se' % name)
+            x = conv_block(x, layout=layout[3:], filters=filters[1:], name='%s-conv2' % name, **kwargs)
+            return x
+        return conv_block(inputs, layout=layout, filters=filters, name=name, **kwargs)
```
roman-kh: To remove complicated implicit configuration, hardcode all the parameters (layout, filters) from the article right in the `head` method.
Later we will add the possibility to chain methods.

batchflow/models/tf/mobilenet.py
dmylzenova: then if there is activation in kwargs, it has to be removed first and added back for the next conv_block.

batchflow/models/tf/mobilenet.py
dmylzenova: in the article and TensorFlow realization it is the same for all layers

dmylzenova: what about complying with a class docstring - there is no activation for the body?

roman-kh: Add it then.
Anyway, by default body is a conv block so it takes all the conv block params

batchflow/models/tf/__init__.py
```diff
@@ -12,7 +12,7 @@
 from .inception_v4 import Inception_v4
 from .inception_resnet_v2 import InceptionResNet_v2
 from .squeezenet import SqueezeNet
-from .mobilenet import MobileNet, MobileNet_v2
+from .mobilenet import MobileNet_v1, MobileNet, MobileNet_v3, MobileNet_v3_small
```
roman-kh: That's quite confusing. The original MobileNet is named MobileNet_v1, while the model named MobileNet is actually MobileNet v2.

batchflow/models/tf/mobilenet.py
```diff
@@ -134,31 +164,43 @@ class MobileNet_v2(TFModel):
...
+                multiplier for the number of filters in internal convolutions
             strides : int
-
+                stride for 3x3 convolution
+            kernel_size : int
+                kernel size for depthwise convolution
+            activation : callable, optional
+            se_block : bool or dict
+                whether to include squeeze and excitation block
+                If dict it must contain ratio and activation keys
```
roman-kh:
```suggestion
                If dict, it must contain :meth:`~TFModel.se_block` params.
```

batchflow/models/tf/mobilenet.py
dmylzenova: hardcoded layout, left filters as parameter modified in build_config.

batchflow/models/tf/__init__.py
dmylzenova: renamed.

batchflow/models/tf/mobilenet.py
dmylzenova: fixed.

batchflow/models/tf/mobilenet.py
dmylzenova: fixed.

batchflow/models/tf/mobilenet.py
dmylzenova: fixed.

batchflow/models/tf/mobilenet.py
dmylzenova: fixed.

batchflow/models/tf/mobilenet.py
dmylzenova: fixed.

batchflow/models/tf/mobilenet.py
dmylzenova: No, it doesn't. Anyway. In this case, the same activation goes everywhere.

batchflow/models/tf/layers/block.py
dmylzenova: done.

batchflow/models/tf/mobilenet.py
dmylzenova: done.

batchflow/models/tf/mobilenet.py
```diff
@@ -158,7 +196,7 @@ def default_config(cls):
 
     def build_config(self, names=None):
         config = super().build_config(names)
-        config['head']['filters'][-1] = self.num_classes('targets')
+        config['head/filters'][-1] = self.num_classes('targets')
```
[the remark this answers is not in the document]
dmylzenova: fixed.

batchflow/models/tf/mobilenet.py
```diff
@@ -112,19 +151,9 @@ def block(cls, inputs, strides=1, double_filters=False, width_factor=1, name=Non
...
-    dict(repeats=1, filters=320, expansion_factor=6, strides=1),
-]
-
-class MobileNet_v2(TFModel):
-    """ MobileNet version 2
-
+class MobileNet(TFModel):
+    """ Base class for MobileNets after v2.
+    Default configuration is MobileNet_v2.
     **Configuration**
```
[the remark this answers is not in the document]
dmylzenova: done.

batchflow/models/tf/mobilenet.py
dmylzenova: fixed

batchflow/models/tf/mobilenet.py
dmylzenova: done.

batchflow/models/tf/base.py
dmylzenova: done.

batchflow/models/tf/mobilenet.py
```diff
@@ -112,19 +152,9 @@ def block(cls, inputs, strides=1, double_filters=False, width_factor=1, name=Non
...
-    dict(repeats=4, filters=64, expansion_factor=6, strides=2),
-    dict(repeats=3, filters=96, expansion_factor=6, strides=1),
-    dict(repeats=3, filters=160, expansion_factor=6, strides=2),
-    dict(repeats=1, filters=320, expansion_factor=6, strides=1),
-]
-
 class MobileNet_v2(TFModel):
-    """ MobileNet version 2
-
+    """ Base class for MobileNets after v2.
```
roman-kh:
```suggestion
    """ MobileNet version 2
```
There is no reference at the top of the file to the article about MobileNet v2

batchflow/models/tf/mobilenet.py
```diff
@@ -112,19 +152,9 @@ def block(cls, inputs, strides=1, double_filters=False, width_factor=1, name=Non
...
+    Default configuration is MobileNet_v2.
```
roman-kh:
```suggestion

```

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +229,114 @@ def body(cls, inputs, name='body', **kwargs):
...
+            x = cls.se_block(x, name='%s-se' % name, **se_block)
+        x = conv_block(x, 'cn', filters, 1, name='%s-down' % name, **kwargs)
+
         if residual:
             x = inputs + x
         return x
+
+
+class MobileNet_v3(MobileNet_v2):
+    """ MobileNet version 3 architecture """
```
roman-kh:
```suggestion
    """ MobileNet version 3 """
```

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +229,114 @@ def body(cls, inputs, name='body', **kwargs):
...
+        config['initial_block'].update(dict(layout='cna', filters=16, kernel_size=3, strides=2, activation=h_swish))
+        config['body'].update(dict(width_factor=1, layout=deepcopy(_V3_SMALL_DEFAULT_BODY)))
+        config['head'].update(dict(layout='cnavcacV', filters=[576, 1280, 2], pool_size=7,
+                                   kernel_size=1, activation=h_swish,
+                                   se_block=dict(activation=[h_swish, h_sigmoid], ratio=144)))
+        config['loss'] = 'ce'
+        return config
+
+    @classmethod
+    def head(cls, inputs, se_block=None, name='head', **kwargs):
```
roman-kh: Move it to MobileNet_v3, as according to the article head differs by config only, not by structure. MobileNet v3 Large could have se block in the head as well

batchflow/models/tf/mobilenet.py
```diff
@@ -189,38 +229,114 @@ def body(cls, inputs, name='body', **kwargs):
...
+        """ The last network layers which produce predictions
+
+        Parameters
+        ----------
+        inputs : tf.Tensor
+            input tensor
+        name : str
+            scope name
+        se_block : dict, optional
+            If not None it must be a dict containing :meth:`~TFModel.se_block` params
```
roman-kh:
```suggestion
            If not None, it must be a dict containing :meth:`~TFModel.se_block` params
```

batchflow/models/tf/mobilenet.py
dmylzenova: there is, the name of the article didn't contain the name of the network

batchflow/models/tf/mobilenet.py
dmylzenova: done.

batchflow/models/tf/mobilenet.py
dmylzenova: done.

batchflow/models/tf/mobilenet.py
dmylzenova: done,

batchflow/models/tf/mobilenet.py
```diff
@@ -189,28 +230,41 @@ def body(cls, inputs, name='body', **kwargs):
                 for k in range(repeats):
                     if k > 0:
                         block['strides'] = 1
-                    x = cls.block(x, **block, residual=k > 0, name='block-%d' % i, **kwargs)
+                    residual = block.pop('residual', False) or k > 0
+                    block = {**kwargs, **block}
+                    x = cls.block(x, **block, residual=residual, name='block-%d' % i)
```
SergeyTsimfer: What about moving `repeat` logic inside block, with `repeat` being one of the `kwargs`? Same goes for `residual`. That would allow to use `body` very similar to the `MobileNet` one.

batchflow/models/tf/mobilenet.py
```diff
@@ -189,28 +230,41 @@ def body(cls, inputs, name='body', **kwargs):
...
+              se_block=False, name=None, **kwargs):
+        """ An inverted residual bottleneck block consisting of a separable depthwise convolution and 1x1 pointise
+        convolution and an optional Squeeze-and-Excitation block.
 
         Parameters
         ----------
         inputs : tf.Tensor
             input tensor
+        repeats : int
+            number of repeats of the block
```
SergeyTsimfer: Strange

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +273,71 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+
+        if se_block:
+            if not isinstance(se_block, dict):
+                se_block = dict(activation=[kwargs.get('activation', tf.nn.relu), h_sigmoid], ratio=num_filters // 4)
+            se_block = {**kwargs, **se_block}
+            x = cls.se_block(x, name='%s-se' % name, **se_block)
+        x = conv_block(x, 'cn', filters, 1, name='%s-down' % name, **kwargs)
+
         if residual:
             x = inputs + x
```
SergeyTsimfer: `TFModel.combine` can be handy

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +273,71 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+        """
+        kwargs = cls.fill_params('head', se_block=se_block, **kwargs)
+        layout, filters, se_block = cls.pop(['layout', 'filters', 'se_block'], kwargs)
+
+        if se_block:
+            x = conv_block(inputs, 'cna', filters[0], name='%s-conv1' % name, **kwargs)
+            x = cls.se_block(x, **{**kwargs, **se_block}, name='%s-se' % name)
+            x = conv_block(x, 'vcacV', filters=filters[1:], name='%s-conv2' % name, **kwargs)
+            return x
+        return conv_block(inputs, layout=layout, filters=filters, name=name, **kwargs)
```
SergeyTsimfer: If `se_block` evaluates to `True`, then `layout` is not used at all. Is that intended?

batchflow/models/tf/mobilenet.py
```diff
@@ -219,8 +273,71 @@ def block(cls, inputs, filters, residual=False, strides=1, expansion_factor=6, w
...
+
+
+class MobileNet_v3(MobileNet_v2):
+    """ MobileNet version 3 large """
+    @classmethod
+    def default_config(cls):
+        config = TFModel.default_config()
+        config['initial_block'].update(dict(layout='cna', filters=16, kernel_size=3, strides=2, activation=h_swish))
+        config['body'].update(dict(width_factor=1, layout=deepcopy(_V3_LARGE_DEFAULT_BODY)))
```
SergeyTsimfer: Looks like you need `deepcopy` because the exact keyword `layout` is used to store parameters, and you need to remove it in order to not interfere with the `layout` parameter of `conv_block`. Maybe it would be easier to change the word to, say, `architecture`, like it is done in `Inception` models?
Besides, there is a difference in formats of `_V1_DEFAULT_BODY ` and the others: the first contains `width_factor` in itself, while all of the latter make it separate parameter for the `body`

batchflow/tests/dataformat_test.py
```diff
@@ -7,10 +7,10 @@
 
 from batchflow import Pipeline, ImagesBatch, Dataset
 from batchflow import B, V, C
-from batchflow.models.tf import VGG7, ResNet18, Inception_v1, MobileNet
+from batchflow.models.tf import VGG7, ResNet18, Inception_v1, MobileNet, MobileNet_v2, MobileNet_v3, MobileNet_v3_small
 
 
-MODELS = [VGG7, ResNet18, Inception_v1, MobileNet]
+MODELS = [VGG7, ResNet18, Inception_v1, MobileNet, MobileNet_v2, MobileNet_v3, MobileNet_v3_small]
```
SergeyTsimfer: What is the time of running the tests on each of these models?
If the time is reasonable to wait, we can add these architectures to `notebooks_tests`.

batchflow/models/tf/mobilenet.py
```diff
@@ -134,31 +161,46 @@ class MobileNet_v2(TFModel):
...
-
         return config
 
     def build_config(self, names=None):
         config = super().build_config(names)
-        config['head']['filters'][-1] = self.num_classes('targets')
+        if isinstance(config['head/filters'], list):
+            config['head/filters'][-1] = self.num_classes('targets')
+        else:
+            config['head/filters'] = self.num_classes('targets')
```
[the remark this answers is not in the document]
roman-kh: I though about this earlier and decided not to change it since this is needed for some models but not for others.
However, now `head` takes `targets` as a parameter so we can move this into `TFModel.head` and other model can easily redefine this behaviour when required.

batchflow/models/tf/mobilenet.py
roman-kh: We decided to leave this the way it is. But multiple `head`s will allow to move this into a model config.

batchflow/models/tf/mobilenet.py
```diff
@@ -310,24 +298,37 @@ def head(cls, inputs, se_block=None, name='head', **kwargs):
         name : str
             scope name
         se_block : dict, optional
-            If not None, it must be a dict containing :meth:`~TFModel.se_block` params
+            params for :meth:`~TFModel.se_block`
```
dmylzenova: It is confusing to see the head with hardcoded parameters from version 3 in version 2.

batchflow/models/tf/mobilenet.py
dmylzenova: Nope.

roman-kh:
@dmylzenova Why did you remove the changes from 277fe4c? Revert them back.

PR #334, opened 2019-07-15
Alter calc origin

dpodvyaznikov:
This PR:
1. Adds new predefined `origin` options in `_calc_origin`;
2. Changes `scale` API for more customisable behaviour when `preserve_shape` is True:
* crop_origin allows to crop from desired location of transformed image;
* input_origin allows to input transformed image in desired point of background;
3. Updates docstrings.


dpodvyaznikov:
I've come up with what seems to be a better way to handle multi-source case, will test it before long.

PR #335, opened 2019-07-18
Torch decay

anton-br:
From torch.optim.lr_scheduler the following decay algorithms have been implemented:
* ExponentialLR
* LambdaLR
* StepLR
* MultiStepLR
* CosineAnnealingLR
* CyclicLR

Moreover new config parameter with name ```n_iters``` were introduced. This parameter should contain the number of iterations in one epoch.

Thus, simple config with decay scheduler can be written this way:
```python
config = {
     ...
    'n_iters': 4,
    'decay': ('exp'),
     ...
}
```

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -36,7 +36,6 @@ DECAYS = {
     'step': torch.optim.lr_scheduler.StepLR,
     'multistep': torch.optim.lr_scheduler.MultiStepLR,
     'cos': torch.optim.lr_scheduler.CosineAnnealingLR,
-    'lronp': torch.optim.lr_scheduler.ReduceLROnPlateau,
     'cyclic': torch.optim.lr_scheduler.CyclicLR,
 
 }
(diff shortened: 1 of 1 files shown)


batchflow/models/torch/base.py
```diff
@@ -72,12 +88,15 @@ class TorchModel(BaseModel):
         - tuple (name, args)
         - dict {'name': name, **args}
 
+        .. note:: All torch decays require to have ```num_iters``` as a key in a configuration
```
roman-kh: n_iters

batchflow/models/torch/base.py
```diff
@@ -338,18 +371,29 @@ def _make_optimizer(self, config):
...
         elif isinstance(decay_name, str) and hasattr(torch.optim.lr_scheduler, decay_name):
             decay_name = getattr(torch.optim.lr_scheduler, decay_name)
         elif decay_name in DECAYS:
-            decay_name = DECAYS.get(re.sub('[-_ ]', '', decay_name).lower(), None)
+            decay_name = DECAYS.get(decay_name)
         else:
             raise ValueError("Unknown learning rate decay method", decay_name)
-
+        if decay_name in DECAYS_DEFAULTS:
+            decay_args = self._update_decay_args(decay_name, decay_args)
```
roman-kh: What's the point in moving 2 lines of code to a separate method?

roman-kh: You don't seem to use anywhere else

roman-kh: In line #385 you update DEFAULT_DICT, while here you update the copy of DEFAULT_DICT, why?

PR #355, opened 2019-07-29
add missing  padding arguments to class construcrots for dowmsample b…

nikita-klsh:
1. In torch models padding parameter for downsample block does not  go down from models config to actual action.
Whenever we call `Pool` layer, `_Pool` class constructor executed twice. First time directly from `Pool`, second time - from one of the `MaxPool / AvgPool` classes.    
During the first execution `padding` pops from `kwargs` therefore the second time `_Pool` constructor is always being called with default padding value.

2. `_get_device()` method:
    1.  The condition ` device is torch.device` is  `False` in case `device = torch.device('cuda')`. I believe it was supposed to return `True`?
    2.  `torch.device('cpu', index)` : unlike CUDA the only available index for CPU device is 0. You can verify it. Its not clear from the documentation whats the difference between `torch.device('cpu')` and `torch.device('cpu', 0)`. However, it becomes critical when we come down to  `load` method, in particular this piece of code:

        `if device: checkpoint = torch.load(path, map_location=device)`

    Works well if `map_location = torch.device('cpu')` and falls with error if  `map_location = torch.device('cpu', 0)`. I suggest to not include index for CPU devices at all.

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -315,14 +315,15 @@ class TorchModel(BaseModel):
 
     def _get_device(self):
         device = self.config.get('device')
-        if device is torch.device or device is None:
+        if isinstance(device, torch.device) or device is None:
             _device = device
         elif isinstance(device, str):
             _device = device.split(':')
             unit, index = _device if len(_device) > 1 else (device, '0')
-            if unit.lower() in ['gpu', 'cpu']:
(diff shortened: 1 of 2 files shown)


roman-kh:
@SergeyTsimfer Should we merge it before #369?

SergeyTsimfer:
> @SergeyTsimfer Should we merge it before #369?

Yes, because I've used this PR as a base for mine

PR #357, opened 2019-07-31
Iss 346 unet

a-arefina:
UNet as EcoderDecoder subclass
+ EncoderDecoder improvements


roman-kh:
Let's move ` config['head/targets'] = self.get_from_attr('targets')` from `build_config` to `TFModel`

batchflow/models/tf/unet.py
```diff
@@ -1,55 +1,59 @@
...
+        config['initial_block'] = dict(layout='cna cna', kernel_size=3, filters=64)
+
+        config['body/encoder/num_stages'] = 4
+        config['body/encoder/blocks'] = dict(layout='cna cna', kernel_size=3, filters=None)
+        config['body/embedding'] = None  # identity
+        config['body/decoder/num_stages'] = None  # defaults to body/encoder/num_stages
+        config['body/decoder/factor'] = None  # defaults to [2] * num_stages
+        config['body/decoder/blocks'] = dict(layout='cna cna', kernel_size=3, filters=None)
+
+        config['head'] = dict(layout='cna cna', kernel_size=[3, 3], strides=1, filters=64)
```
roman-kh: Why not just 3?

a-arefina: Legacy. I'll fix it

a-arefina:
> Let's move ` config['head/targets'] = self.get_from_attr('targets')` from `build_config` to `TFModel`

'head/targets' is not needed in all `TFModel`s, only in `EncoderDecoder` and its children

roman-kh:
> > Let's move ` config['head/targets'] = self.get_from_attr('targets')` from `build_config` to `TFModel`
> 'head/targets' is not needed in all `TFModel`s, only in `EncoderDecoder` and its children
That's right. But some models might use it. So having the same method signature would be great.

batchflow/models/tf/encoder_decoder.py
```diff
@@ -66,6 +66,10 @@ class EncoderDecoder(TFModel):
                 other args : dict
                     Parameters for the base block.
 
+    head : dict
+        parameters for the head layers, usually :func:`.conv_block` parameters
+        an extra 1x1 convolutional layer that makes outputs shape match targets will be added implicitly if needed
```
SergeyTsimfer:
```suggestion
        Note that an extra 1x1 convolution may be applied in order to make predictions compatible with the shape of targets
```

batchflow/models/tf/encoder_decoder.py
```diff
@@ -146,11 +152,12 @@ def body(cls, inputs, name='body', **kwargs):
...
-            embeddings = embeddings if isinstance(embeddings, (tuple, list)) else [embeddings]
+            if embeddings is not None:
+                embeddings = embeddings if isinstance(embeddings, (tuple, list)) else [embeddings]
 
-            for i, embedding in enumerate(embeddings):
-                embedding_args = {**kwargs, **embedding}
-                x = cls.embedding(x, name='embedding'+str(i), **embedding_args)
+                for i, embedding in enumerate(embeddings):
+                    embedding_args = {**kwargs, **embedding}
+                    x = cls.embedding(x, name='embedding'+str(i), **embedding_args)
```
SergeyTsimfer:
```suggestion
                    x = cls.embedding(x, name='embedding-'+str(i), **embedding_args)
```

batchflow/models/tf/encoder_decoder.py
```diff
@@ -163,10 +170,16 @@ def body(cls, inputs, name='body', **kwargs):
     def head(cls, inputs, targets, name='head', **kwargs):
         """ Linear convolutions. """
         kwargs = cls.fill_params('head', **kwargs)
+
         with tf.variable_scope(name):
             x = cls.crop(inputs, targets, kwargs['data_format'])
+            x = super().head(x, name, **kwargs)
+
```
SergeyTsimfer: You don't need extra scope named `name` inside already named `name` scope

a-arefina: Couldn't find better name, so leaving this as is

batchflow/models/tf/encoder_decoder.py
```diff
@@ -20,16 +20,16 @@ class EncoderDecoder(TFModel):
         encoder : dict
             base : TFModel
                 Model implementing ``make_encoder`` method which returns tensors
-                with encoded representation of the inputs.
+                with encoded representation of the inputs. Optional
```
SergeyTsimfer: move `optional` keyword to the first line of parameter description, e.g. `base : TFModel, optional`, just like [numpy style](https://numpydoc.readthedocs.io/en/latest/format.html) advises

batchflow/models/tf/encoder_decoder.py
```diff
@@ -121,15 +130,11 @@ def default_config(cls):
...
-        config['body/decoder/blocks'] = dict(base=cls.block, combine_op='softsum')
-        config['head'] = dict(layout='c', kernel_size=1)
-        return config
+        config['body/decoder/blocks'] = dict(base=cls.block, combine_op='concat')
 
-    def build_config(self, names=None):
-        config = super().build_config(names)
-        config['head/targets'] = self.get_from_attr('targets')
-        return config
+        config['head/layout'] = None
```
SergeyTsimfer: is this line necessary?

a-arefina: This is to explicitly show what the default parameter is

roman-kh:
This closes #346

PR #370, opened 2019-08-12
Iss 246 model load save test

a-arefina:
* added test  for model loading and saving
* moved fixture for dataset and model config creation to batchflow/tests/conftest.py


batchflow/tests/conftest.py
```diff
@@ -1,5 +1,36 @@
...
+def model_setup_images_clf():
+    """ Pytest fixture to generate fake dataset and model config for image classification
+
+    Returns
+    -------
+    tuple
+        an instance of Dataset
+        a model config
+    """
+    def _model_setup(dataset_size=50, image_shape=(2, 100, 100), num_classes=10):
```
SergeyTsimfer: You don't seem to use `dataset_size` and `num_classes` parameters anywhere. Are they necessary?

SergeyTsimfer: You can add `data_format` parameter to transpose `image_shape`. That would allow to use this fixture in `dataformat_test` as well.

SergeyTsimfer: Actually, that would remove the need for `image_shape` parameter as well

batchflow/tests/model_save_load_test.py
```diff
@@ -0,0 +1,177 @@
...
+from batchflow.models.torch import VGG7 as TORCH_VGG7
+
+PATH = 'my_mdl'
+BATCH_SIZE = 20
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize(
+    'mode',
+    [pytest.param('static', marks=pytest.mark.skip(reason="doesn't work")),
```
SergeyTsimfer: Is it not working at all?

batchflow/tests/model_save_load_test.py
```diff
@@ -0,0 +1,177 @@
...
+
+        def make_path_name(batch, path, iteration):
+            _ = batch
+            return "{}_{}".format(path, iteration)
+
+        save_tmpl = (Pipeline()
+                     .init_variable('save_path')
+                     .call(make_path_name, save_to=V("save_path"),
+                           kwargs=dict(path=save_path, iteration=I("current")))
+                     .save_model('model', path=V("save_path"))
```
SergeyTsimfer: Can't you just `path = save_path + I('current')`?

batchflow/tests/model_save_load_test.py
```diff
@@ -0,0 +1,177 @@
...
+        return save_pipeline, load_pipeline
+
+    @pytest.fixture
+    def fixture_dict(self, tf_pipelines, torch_pipelines):
+        """
+        workaround for passing parametrized fixtures as parameters
+        """
+        return dict(tf_pipelines=tf_pipelines, torch_pipelines=torch_pipelines)
+
+    @pytest.mark.parametrize('fixture_name, model_args, model_kwargs',
```
SergeyTsimfer: `model_args` and `model_kwargs` are used to take into account different order of parameters for `tf`/`torch` versions of `train`. Why can't you use the same logic for `predict`? That would make `tf_pipeline` and `torch_pipeline` look very similar. The only real difference would be in `to_array` call, and you can easily avoid that with pipeline config

batchflow/tests/model_save_load_test.py
```diff
@@ -0,0 +1,177 @@
...
+
+        save_pipeline.after.save_model('model', path=save_path)
+        save_pipeline.run(BATCH_SIZE, n_epochs=1)
+        saved_predictions = save_pipeline.get_variable('predictions')
+
+        load_pipeline.before.load_model(mode, C('model_class'), 'model', path=save_path)
+        load_pipeline.run(BATCH_SIZE, n_epochs=1)
+        loaded_predictions = load_pipeline.get_variable('predictions')
+
+        assert (np.concatenate(saved_predictions) == np.concatenate(loaded_predictions)).all()
```
SergeyTsimfer: All the pipeline interactions are tested. However, if it is `Pipeline` that is broken, and `TFModel`/`TorchModel` are working fine, all of these tests still would fail. Can we add `pipeline`-less test to account for this situation?

batchflow/tests/model_save_load_test.py
a-arefina: saving static models is broken, i'm investigating it right now
I will split mode parametrization for load and save pipelines

batchflow/tests/model_save_load_test.py
a-arefina: I thought about it, but we have to parametrize predict and train separately because in torch we have different sets of positional args in train and in predict, so I thought it would be over complicated. 
I'll try to make all parametrization in config and compare with current state

batchflow/tests/model_save_load_test.py
a-arefina: removed `'static'`  after discussion with @roman-kh

PR #373, opened 2019-08-16
Rebatch test

a-arefina:
2 fixes for rebatch + test

### DIFF
--- a/batchflow/batch.py
+++ b/batchflow/batch.py
@@ -147,7 +147,6 @@ class Batch:
 
         if batch_size is None:
             break_point = len(batches)
-            last_batch_len = len(batches[-1])
         else:
             break_point = len(batches) - 1
             last_batch_len = 0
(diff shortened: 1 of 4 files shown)


batchflow/batch.py
```diff
@@ -175,14 +174,14 @@ def _make_batch(data):
             if batch_size is None:
                 new_comp = [b.get(component=comp) for b in batches[:break_point]]
             else:
-                b = batches[break_point]
-                last_batch_len_ = b.get_pos(None, comp, b.indices[last_batch_len - 1])
+                last_batch = batches[break_point]
+                last_batch_last_index = last_batch.get_pos(None, comp, last_batch.indices[last_batch_len - 1])
                 new_comp = [b.get(component=comp) for b in batches[:break_point]] + \
```
a-arefina: these are just readability improvements to get rid of variable `b` both inside and outside of list comprehensions

batchflow/batch.py
```diff
@@ -149,7 +149,7 @@ def _make_batch(data):
             break_point = len(batches)
             last_batch_len = len(batches[-1])
```
dpodvyaznikov: I think this line can be removed

PR #375, opened 2019-08-19
Linknet update

Dimonovez:
Linknet scheme is the same as in the article

### DIFF
--- a/batchflow/models/tf/encoder_decoder.py
+++ b/batchflow/models/tf/encoder_decoder.py
@@ -169,12 +169,12 @@ class EncoderDecoder(TFModel):
         kwargs = cls.fill_params('head', **kwargs)
 
         with tf.variable_scope(name):
-            x = cls.crop(inputs, targets, kwargs['data_format'])
-            x = super().head(x, name, **kwargs)
+            x = super().head(inputs, name, **kwargs)
+            x = cls.crop(x, targets, kwargs['data_format'])
 
             channels = cls.num_channels(targets)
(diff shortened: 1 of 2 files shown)


batchflow/models/tf/encoder_decoder.py
```diff
@@ -169,12 +169,12 @@ def head(cls, inputs, targets, name='head', **kwargs):
         kwargs = cls.fill_params('head', **kwargs)
 
         with tf.variable_scope(name):
-            x = cls.crop(inputs, targets, kwargs['data_format'])
-            x = super().head(x, name, **kwargs)
+            x = super().head(inputs, name, **kwargs)
+            x = cls.crop(x, targets, kwargs['data_format'])
```
SergeyTsimfer: Lets make a check: if spatial shape of `inputs` is bigger than that of `targets`, we `crop`. Then we call `head`, then we `crop` once again

batchflow/models/tf/encoder_decoder.py
```diff
@@ -167,10 +167,13 @@ def body(cls, inputs, name='body', **kwargs):
     def head(cls, inputs, targets, name='head', **kwargs):
         """ Linear convolutions. """
         kwargs = cls.fill_params('head', **kwargs)
+        data_format = kwargs['data_format']
 
         with tf.variable_scope(name):
+            if cls.spatial_shape(inputs, data_format) > cls.spatial_shape(targets, data_format):
```
roman-kh: How would this work with dynamic shapes?

Issue #364, opened 2019-08-06
Update conv_block layers


roman-kh:
- rename DropBlock to B
- get back AlphaDropout as D
- add depth-wise only separable conv
- reconcile local and tf separable layers

roman-kh:
Incorporate #396 as well

roman-kh:
@SergeyTsimfer Is everything updated?

SergeyTsimfer:
Yes

PR #376, opened 2019-08-19
TFourteen

SergeyTsimfer:
This PR proposes to:

-  change `tensorflow` layers to their `keras` counterparts

-  change signatures of `conv_block` and underlying functions to look more like modern `tf` API: now it is a class `ConvBlock` with  overloaded `__call__` method and must be called as`ConvBlock(*args, **kwargs)(input)`

- `ConvBlock.add_letter` method allows to control `layout` parsing from the outside

- old syntax is preserved via `add_as_function` decorator


batchflow/models/tf/layers/utils.py
```diff
@@ -0,0 +1,22 @@
+""" Common utility functions for all of TensorFlow layers. """
+import inspect
+
+
+
+def add_as_function(cls):
+    """ Decorator for classes. Automatically adds functional interface for `call` method of class. """
+    name = cls.__name__
+    func_name = ''.join('_' + c.lower()
+                        if (c.isupper() and (i != len(name)-1) and name[i+1].islower()) else c.lower()
```
a-arefina: c.lower() if ... else c.lower() ???

batchflow/models/tf/layers/core.py
```diff
@@ -2,78 +2,101 @@
...
-def flatten(inputs, name=None):
-    """ Flatten tensor to two dimensions (batch_size, item_vector_size) using inferred shape and numpy """
-    x = tf.convert_to_tensor(inputs)
-    shape = x.get_shape().as_list()
-    dim = np.prod(shape[1:])
-    x = tf.reshape(x, [-1, dim], name=name)
-    return x
 
 
-def alpha_dropout(inputs, rate=0.5, noise_shape=None, seed=None, training=False, name=None):
```
a-arefina: Did you remove this functionality completely?

batchflow/models/tf/layers/conv_block.py
```diff
@@ -1,44 +1,31 @@
 """ Contains convolution layers """
+import logging
 import numpy as np
 import tensorflow as tf
+import tensorflow.keras.layers as K #pylint: disable=import-error
```
a-arefina: There is no common style for this alias but I would rather use something in lower case and not a single letter, `kl`, for example

batchflow/models/tf/layers/utils.py
Dimonovez: Join  `'_' + c.lower()` or `c.lower()`

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+        .replace('V', 'P')
+        .replace('D', 'd')
+        .replace('O', 'd')
+        .replace('A', 'b')
+        .replace('B', 'b')
+        .replace('N', 'b')
+        .replace('X', 'b')
+    )
+
+    C_GROUPS = dict(zip(LAYER_KEYS, GROUP_KEYS))
```
a-arefina:
```suggestion:-16+14
    C_GROUPS = {k: k for k in C_LAYERS}
    C_GROUPS.update({
        't': 'c',
        'C': 'c',
        'T': 'c',
        'v': 'p',
        'V': 'P',
        'D': 'd',
        'O': 'd',
        'A': 'b',
        'B': 'b',
        'N': 'b',
        'X': 'b',
    })
```

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+
+        context = None
+        if self.name is not None:
+            context = tf.variable_scope(self.name, reuse=self.kwargs.get('reuse'))
+            context.__enter__()
+
+        layout_dict = {}
+        for layer in layout:
+            if self.C_GROUPS[layer] not in layout_dict:
+                layout_dict[self.C_GROUPS[layer]] = [-1, 0]
```
a-arefina: layout_dict.setdefault(self.C_GROUPS[layer], [-1, 0])

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+            layer_name = self.C_LAYERS[layer]
+            layer_fn = self.FUNC_LAYERS[layer_name]
+
+            args = {}
+            call_args = {}
+
+            if layer == 'a':
+                args = dict(activation=self.activation)
+                layer_fn = unpack_args(args, *layout_dict[self.C_GROUPS[layer]])['activation']
+                if layer_fn is not None:
```
a-arefina: How can it be not None?

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+                    tensor = layer_fn(tensor)
+            elif layer == 'R':
+                residuals += [tensor]
+            elif layer == 'A':
+                args = dict(factor=self.kwargs.get('factor'), data_format=self.data_format)
+                args = unpack_args(args, *layout_dict[self.C_GROUPS[layer]])
+                t = self.FUNC_LAYERS['resize_bilinear_additive'](tensor, **args, name='rba-%d' % i)
```
a-arefina: Maybe we should just set `'residual_bilinear_additive': ResizeBilinearAdditive`, in `FUNC_LAYERS` with proper comment and call `layer_fn` here?

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+            layout_dict[self.C_GROUPS[layer]][0] += 1
+            layer_name = self.C_LAYERS[layer]
+            layer_fn = self.FUNC_LAYERS[layer_name]
+
+            args = {}
+            call_args = {}
+
+            if layer == 'a':
+                args = dict(activation=self.activation)
+                layer_fn = unpack_args(args, *layout_dict[self.C_GROUPS[layer]])['activation']
```
a-arefina: refactoring: 
1) add `layout_dict[self.C_GROUPS[layer]]` as a variable, because it is used multiple times
2) rename `layer_fn` to `activation_fn` or something like this, because now it shadows upper `layer_fn`

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+                if layer_fn is not None:
+                    tensor = layer_fn(tensor)
```
a-arefina: add warning otherwise?

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+                    if units is None:
+                        raise ValueError('units cannot be None if layout includes dense layers')
+                    args = dict(units=units)
+
+                elif layer == 'c':
+                    args = dict(filters=self.filters, kernel_size=self.kernel_size, strides=self.strides,
+                                padding=self.padding, data_format=self.data_format,
+                                dilation_rate=self.dilation_rate)
+                    if self.filters is None or self.filters == 0:
+                        raise ValueError('filters cannot be None or 0 if layout includes convolutional layers')
```
a-arefina: What about making these 2 lines a function `check_filters_for_convolution`?

batchflow/models/tf/layers/conv_block.py
```diff
@@ -159,33 +143,296 @@ def conv_block(inputs, layout='', filters=0, kernel_size=3, name=None,
...
+                    args = dict(depth=self.kwargs.get('depth'), data_format=self.data_format)
+
+                elif layer in ['b', 'B', 'N', 'X']:
+                    args = dict(factor=self.kwargs.get('factor'),
+                                shape=self.kwargs.get('shape'),
+                                data_format=self.data_format)
+                    if self.kwargs.get('upsampling_layout'):
+                        args['layout'] = self.kwargs.get('upsampling_layout')
+
+                else:
```
a-arefina: elif layer not in self.C_LAYERS.keys():
raise ValueError('Unknown layer symbol - %s' % layer)

batchflow/models/tf/layers/conv.py
```diff
@@ -1,199 +1,211 @@
...
-        kernel size
+        Kernel size.
     strides : int
-        convolution stride. Default is 1.
+        Convolution stride. Default is 1.
+    """
+    def __init__(self, filters, kernel_size, strides=1, padding='valid', data_format='channels_last', *args, **kwargs):
+        self.filters, self.kernel_size, self.strides = filters, kernel_size, strides
+        self.padding, self.data_format = padding, data_format
+        self.args, self.kwargs = args, kwargs
```
a-arefina: I feel very uneasy about converting all these parameters from fields to arguments in ConvBlock and vice versa here. If we need to change parameters list we'll have to check 4 places: ConvBlock init, inside case branch, here and in `__call__ `

What about introducing a small base class
```
class ParameterBuilder:
    parameters = []

    def __init__(self, *args, **kwargs):
        for p in self.parameters:
            self.__setattr__(p, kwargs.pop(p))

        self.args, self.kwargs = args, kwargs
```
we could use it like this:
```
class Conv1DTransposeNn(ParameterBuilder):
    parameters = ['filters', 'output_shape', 'strides', 'padding', 'data_format']

    def my_print(self):
        print(self.padding)
```

and in ConvBlock we could assemble parameters like this:
```
args = {p: getattr(self, p) for p in layer_fn.parameters if hasattr(self, p)}
```

and we will perhaps have to rename `layer_fn` to `layer_cls` in ConvBlock to make it more clear

batchflow/models/tf/layers/conv.py
```diff
@@ -1,199 +1,211 @@
...
+
+    def __call__(self, inputs):
+        axis = 1 if self.data_format == 'NWC' else 2
+        inputs = tf.expand_dims(inputs, axis=axis)
+        filters = tf.expand_dims(self.filters, axis=0)
+        output_shape = tf.concat([self.output_shape[:axis], (1, ), self.output_shape[axis:]], axis=-1)
+        strides = self.strides[:axis] + [1] + self.strides[axis:]
+        x = tf.nn.conv2d_transpose(inputs, filters, output_shape, strides,
```
a-arefina: Do you need to call `Conv2DTranspose` here?

batchflow/models/tf/layers/drop_block.py
```diff
@@ -32,11 +31,21 @@ def dropblock(inputs, dropout_rate, block_size, is_training, data_format, global
...
+        self.kwargs = kwargs
+
+    def __call__(self, inputs, training):
+        return dropblock(inputs, dropout_rate=self.dropout_rate, block_size=self.block_size,
+                         is_training=training, data_format=self.data_format,
+                         global_step=self.global_step, seed=self.seed, **self.kwargs)
+
+
+
+def dropblock(inputs, dropout_rate, block_size, is_training, data_format, global_step=None, seed=None, **kwargs):
```
a-arefina: Do you need it as separate function? If yes, why not `@add_as_function`?

batchflow/models/tf/layers/pyramid.py
```diff
@@ -2,40 +2,43 @@
...
+        Layer name that will be used as a scope.
     """
+    def __init__(self, *args, **kwargs):
+        self.args, self.kwargs = args, kwargs
+
+    def __call__(self, inputs):
+        return pyramid_pooling(inputs, *self.args, **self.kwargs)
+
+
+def pyramid_pooling(inputs, layout='cna', filters=None, kernel_size=1, pool_op='mean', pyramid=(0, 1, 2, 3, 6),
```
a-arefina: same question as https://github.com/analysiscenter/batchflow/pull/376/files#r317103007

batchflow/models/tf/layers/resize.py
```diff
@@ -100,7 +98,7 @@ def _depth_to_space(inputs, block_size, name='d2s'):
             fltr = np.stack(fltr, axis=-1)
             fltr = np.transpose(fltr, axes=list(range(dim))+[dim, dim+1])
             fltr = tf.constant(fltr, tf.float32)
-            x = conv_layer(inputs, fltr, output_shape, [1] + [block_size] * dim + [1])
+            x = conv_layer(fltr, output_shape, [1] + [block_size] * dim + [1])(inputs)
```
a-arefina:
```suggestion
            x = ConvTranspose(fltr, output_shape, [1] + [block_size] * dim + [1])(inputs)
```

batchflow/models/tf/layers/resize.py
```diff
@@ -65,15 +69,9 @@ def depth_to_space(inputs, block_size, name='d2s', data_format='channels_last'):
...
-
 def _depth_to_space(inputs, block_size, name='d2s'):
     dim = inputs.shape.ndims - 2
-    if dim == 1:
-        conv_layer = conv1d_transpose_nn
-    elif dim == 2:
-        conv_layer = tf.nn.conv2d_transpose
-    elif dim == 3:
-        conv_layer = tf.nn.conv3d_transpose
+    conv_layer = ConvTranspose
```
a-arefina:
```suggestion
```

batchflow/models/tf/layers/resize.py
```diff
@@ -141,38 +143,42 @@ def subpixel_conv(inputs, factor=2, name='subpixel', data_format='channels_last'
     x = inputs
     with tf.variable_scope(name):
         if layout:
-            x = conv_block(inputs, layout, kernel_size=1, name='conv', data_format=data_format, **kwargs)
+            from .conv_block import ConvBlock
```
a-arefina: move this to file beginning

batchflow/models/tf/layers/resize.py
```diff
@@ -141,38 +143,42 @@ def subpixel_conv(inputs, factor=2, name='subpixel', data_format='channels_last'
...
+    def __call__(self, inputs):
+        return resize_bilinear_additive(inputs, *self.args, **self.kwargs)
+
+def resize_bilinear_additive(inputs, factor=2, name='bilinear_additive', data_format='channels_last', **kwargs):
+    """ Resize input tensor with bilinear additive technique. """
     dim = inputs.shape.ndims - 2
     _, channels = _calc_size(inputs, factor, data_format)
     layout = kwargs.pop('layout', 'cna')
     with tf.variable_scope(name):
+        from .conv_block import ConvBlock
```
a-arefina: to file beginning

batchflow/models/tf/layers/utils.py
a-arefina: `''.join(c if c.islower() else ('_' + c.lower()) for c in name).lstrip('_')`

batchflow/models/tf/deeplab.py
```diff
@@ -39,8 +39,6 @@ def default_config(cls):
         config['body/embedding'] = dict(base=aspp)
 
         config['body/decoder/upsample'] = dict(layout='b')
-
-        config['head'] = dict(layout='c', kernel_size=1)
```
a-arefina: please comment on this

batchflow/models/tf/layers/resize.py
a-arefina: @roman-kh told me about recursive import here, but we definitely need some comment in code
on this
And perhaps some layout control so that no one could pass 'B' in layout here

batchflow/models/tf/layers/core.py
SergeyTsimfer: It is present in `Keras`. Restored and assigned to `D` letter in `layout`-convention

batchflow/models/tf/layers/conv_block.py
SergeyTsimfer: Moved it to `conv`-related classes. I can move it to `ConvLayer` class and inherit all of the `conv` related classes from it, but that would not be looking all that nice. Will try though

batchflow/models/tf/layers/conv_block.py
SergeyTsimfer: Changed

batchflow/models/tf/layers/conv.py
SergeyTsimfer: I achieved quite similar (in some ways) functional via `Layer` class, though it is not similar (in some other ways) to the class you described.
You explicitly define list of parameters and then build `__init__` based on this list. This approach lacks defaults for parameters (which can be quite easily added), and it would be hard to add some additional checks in `__init__` (like checking that `conv`-filters are positive ints).
My approach is to define `__init__` explicitly and create list of parameters via `property`. I think that explicit `__init__` helps while looking at class definitions, although your suggestion is great too

batchflow/models/tf/layers/drop_block.py
SergeyTsimfer: I can move it inside `__call__`, but that can wait for other things to be polished

batchflow/models/tf/layers/resize.py
SergeyTsimfer: Done

batchflow/models/tf/layers/resize.py
SergeyTsimfer: Added comment

batchflow/models/tf/deeplab.py
SergeyTsimfer: Previously, it was OK to have non-empty layout here. Now it is passed to `TFModel.head` directly and thus needs `filters`. This parameters (layout and kernel size) are now part of `EncoderDecoder.head`

batchflow/models/tf/layers/conv_block.py
SergeyTsimfer: `linear` activation is OK too

batchflow/models/tf/layers/conv_block.py
SergeyTsimfer: Is that necessary?

batchflow/models/tf/layers/conv_block.py
SergeyTsimfer: Probably the best way is to directly import needed layers like `Conv2D` from `tensorflow.keras.layers`

batchflow/models/tf/layers/conv_block.py
a-arefina: No, it's up to you. IMO my proposal is just more clear (and 2 lines shorter).

batchflow/models/tf/layers/conv.py
SergeyTsimfer: Nope, that is a function that operates directly with `filters`-`Tensor`, not with shapes

batchflow/models/tf/layers/conv_block.py
SergeyTsimfer: Done

batchflow/models/tf/layers/conv_block.py
SergeyTsimfer: Done

batchflow/models/tf/base.py
```diff
@@ -957,6 +957,11 @@ def get_tensor_config(self, tensor, **kwargs):
         config = {**config, **kwargs}
         return config
 
+    def _set_learning_phase(self, mode):
```
roman-kh: Why not `eval`?

SergeyTsimfer: Why it should be call `eval`, considering this function does not evaluate anything?

roman-kh: Like in pytorch - https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.eval

SergeyTsimfer: To stay consistent with pytorch we should add two functions: `train` to set learning phase to train mode, `eval` to set learning phase to inference

roman-kh: `eval(True)` and `eval(False)` can do the trick

SergeyTsimfer:
This PR also addresses #364:

- `DropBlock` is renamed to `O`

- `AlphaDropout` is available as `D`

- adds two depth-wise layers: `w` for the usual one and `W` for the transposed

- swap (where possible) our custom layers to their `keras` counterparts, at least for some of the dimensions. _Most of this was actually done in #328_

PR #380, opened 2019-08-23
Introduce SCSE block

dpodvyaznikov:
This PR proposes to add `scse_block` class method to TFModel.

### DIFF
--- a/batchflow/models/tf/base.py
+++ b/batchflow/models/tf/base.py
@@ -2090,7 +2090,6 @@ class TFModel(BaseModel):
         -------
         tf.Tensor
         """
-
         with tf.variable_scope(name):
             cse = cls.se_block(inputs, ratio, name='cse', **kwargs)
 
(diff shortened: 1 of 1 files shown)


batchflow/models/tf/base.py
```diff
@@ -2071,6 +2071,37 @@ def se_block(cls, inputs, ratio, name='se', **kwargs):
...
+        """
+
+        with tf.variable_scope(name):
+            cse = cls.se_block(inputs, ratio, name='cse', **kwargs)
+
+            x = conv_block(inputs, **{**kwargs, 'layout': 'ca', 'filters': 1, 'kernel_size': 1,
+                                      'activation': tf.nn.sigmoid, 'name': 'sse'})
+
+            sse = tf.multiply(x, inputs)
+            scse = cse + sse
```
roman-kh: Why not just one-liner?

dpodvyaznikov: Used it for testing, forgot to remove.

PR #381, opened 2019-08-26
new_opensets

nikita-klsh:
This PR proposes to add `Imagenette` and `Imagewoof` opensets  and its versions to `batchflow/opensets`

### DIFF
--- a/batchflow/opensets/small_imagenet.py
+++ b/batchflow/opensets/small_imagenet.py
@@ -20,7 +20,7 @@ from .. import DatasetIndex
 logger = logging.getLogger('SmallImagenet')
 
 
-class SmallImagenet(ImagesOpenset):
+class Imagenette(ImagesOpenset):
     """ Base class for ImagenetLike datasets
 
     Notes
@@ -29,7 +29,7 @@ class SmallImagenet(ImagesOpenset):
       Argument `drop_grayscale` controls whether grayscale images should be dropped.
 
     """
-    SOURCE_URL = None
+    SOURCE_URL = 'https://s3.amazonaws.com/fast-ai-imageclas/imagenette.tgz'
     num_classes = 10
 
     def __init__(self, *args, drop_grayscale=True, bar=False, preloaded=None, train_test=True, **kwargs):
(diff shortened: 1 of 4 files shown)


batchflow/opensets/small_imagenet.py
```diff
@@ -0,0 +1,124 @@
...
+
+logger = logging.getLogger('SmallImagenet')
+
+
+class SmallImagenet(ImagesOpenset):
+    """ Base class for ImagenetLike datasets """
+    SOURCE_URL = None
+
+    def __init__(self, *args, bar=False, preloaded=None, train_test=True, **kwargs):
+        self.bar = tqdm.tqdm(total=3) if  bar else None
```
roman-kh: Why `3` whereas there are only two `bar.update()`?

nikita-klsh:
1. Control grayscale images load into dataset.
2. Updat `to_array` ImageBatch action to work with `3 channels` images

batchflow/opensets/small_imagenet.py
```diff
@@ -0,0 +1,138 @@
...
+            return PIL.Image.open(BytesIO(data))
+
+        def _gather_extracted(archive, files):
+            images = np.array([_extract(archive, file) for file in files], dtype=object)
+            labels = np.array([_image_class(file.name) for file in files])
+            labels_encoded = LabelEncoder().fit_transform(labels)
+
+            if self.drop_grayscale:
+                rgb_mask = [i for i in range(len(images)) if images[i].mode == 'RGB']
+                return images[rgb_mask], labels_encoded[rgb_mask]
```
roman-kh: This requires a full copy of data which takes a lot of time and resources.
Another problem is that when they fix grayscale images this becomes an awful resource waste with no reason whatsoever.

batchflow/opensets/small_imagenet.py
```diff
@@ -0,0 +1,138 @@
...
+        logger.info("Extracted")
+        if self.bar:
+            self.bar.update(1)
+
+        self._train_index = DatasetIndex(np.arange(len(train_data[0])))
+        self._test_index = DatasetIndex(np.arange(len(test_data[0])))
+        return train_data, test_data
+
+
+class Imagenette(SmallImagenet):
```
roman-kh: It seems that SmallImagenet is not needed at all. Just rename SmallImagenet to Imagenette and you'll reduce complexity of the code.

batchflow/opensets/small_imagenet.py
```diff
@@ -0,0 +1,136 @@
+""" Contains Imagenette and Imagewoof datasets """
```
SergeyTsimfer: We can also rename filename

batchflow/opensets/small_imagenet.py
```diff
@@ -0,0 +1,136 @@
...
+            self.bar.update(1)
+
+        self._train_index = DatasetIndex(np.arange(len(train_data[0])))
+        self._test_index = DatasetIndex(np.arange(len(test_data[0])))
+        return train_data, test_data
+
+
+class Imagenette320(Imagenette):
+    """ The '320px' version of Imagenette.
+    The shortest size resized to that size with their aspect ratio maintained.
```
SergeyTsimfer: Can we add some info about datasets in the docstring? Knowing `size` would be really nice

PR #394, opened 2019-08-30
Crop origin

dpodvyaznikov:


batchflow/batch_image.py
```diff
@@ -439,6 +441,11 @@ def _calc_origin(self, image_shape, origin, background_shape):
             elif origin == 'random':
                 origin = (np.random.randint(background_shape[0]-image_shape[0]+1),
                           np.random.randint(background_shape[1]-image_shape[1]+1))
+        elif all(0 <= elem < 1 for elem in origin):
+            region = ((background_shape[0]-image_shape[0]+1),
+                      (background_shape[1]-image_shape[1]+1))
+            origin = np.asarray(origin) * region
+
```
roman-kh: what will happen otherwise?

dpodvyaznikov: Method's behavior is the same as it was before in all other cases

roman-kh: If you can improve it easily, there is no point in leaving it without changes.
So it would be nice to throw exceptions if the origin str is not one of those predefined, and it contains a mix of ints and floats

Issue #396, opened 2019-08-31
Add multi-sample dropout


roman-kh:
https://arxiv.org/abs/1905.09788

for `tf` and `torch`

roman-kh:
@SergeyTsimfer What about `torch`?

SergeyTsimfer:
> @SergeyTsimfer What about `torch`?

It was added to `torch` as a part of #438

PR #398, opened 2019-09-03
Multi-sample dropout

SergeyTsimfer:
This PR addresses #396 for `tf`. It also adds capability to work with functional `dropout_rate` both for `Dropout` and `AlphaDropout`.


batchflow/models/tf/layers/core.py
```diff
@@ -50,25 +50,71 @@ def __call__(self, inputs):
...
 
-class Dropout(Layer):
-    """ Wrapper for dropout layer. """
-    def __init__(self, dropout_rate, **kwargs):
+class BaseDropout(Layer):
+    """ Base class for dropout layers.
+
+    Parameters
+    ----------
+    dropout_prob : float, tf.Tensor, callable
```
roman-kh:
```suggestion
    dropout_rate : float, tf.Tensor, callable
```

batchflow/models/tf/layers/core.py
```diff
@@ -50,25 +50,71 @@ def __call__(self, inputs):
...
+        If float or Tensor, then fraction of the input units to drop.
+        If callable, then function to be called on `global_step`. Must return tensor of size 1.
+    multisample: bool, number, sequence
```
roman-kh: Add an empty line before

batchflow/models/tf/layers/core.py
```diff
@@ -50,25 +50,71 @@ def __call__(self, inputs):
...
+    """
+    def __init__(self, dropout_rate, multisample=False, global_step=None, **kwargs):
         self.dropout_rate = dropout_rate
+        self.global_step = global_step
+        self.multisample = multisample
         self.kwargs = kwargs
 
     def __call__(self, inputs, training):
-        return K.Dropout(rate=self.dropout_rate, **self.kwargs)(inputs, training)
+        #pylint: disable=singleton-comparison
```
roman-kh: Remove

batchflow/models/tf/layers/core.py
```diff
@@ -50,25 +50,71 @@ def __call__(self, inputs):
...
+                if all([isinstance(item, int) for item in self.multisample]):
+                    sizes = self.multisample
+                elif all([isinstance(item, float) for item in self.multisample]):
+                    batch_size = tf.cast(tf.shape(inputs)[0], dtype=tf.float32)
+                    sizes = tf.convert_to_tensor([batch_size*item for item in self.multisample])
+                    sizes = tf.cast(tf.math.round(sizes), dtype=tf.int32)
+            elif isinstance(self.multisample, tf.Tensor):
+                sizes = self.multisample
+
+            splitted = tf.split(inputs, sizes, axis=0, name='mdropout_split')
```
roman-kh: The capability to route different batch items to different dropouts is great.
But I reckon we also need the whole batch to go through each dropout, and after that combine is needed.

dpodvyaznikov:
I'd suggest to move check for `dropout_rate==0.0` from `tf.cond` to `if` statement in `__call__`  method of `Dropblock` class.
This way `tf.cond` would only be responsible for train / inference regime switch.

batchflow/models/tf/layers/core.py
```diff
@@ -50,25 +51,121 @@ def __call__(self, inputs):
...
+
+        if self.multisample is not False:
+            if self.multisample is True:
+                self.multisample = 2
+            elif isinstance(self.multisample, float):
+                self.multisample = [self.multisample, 1 - self.multisample]
+
+            if isinstance(self.multisample, int): # dropout to the whole batch, then average
+                dropped = [d_layer(inputs, training) for _ in range(self.multisample)]
+                output = Combine(op='avg')(dropped)
```
dpodvyaznikov: I think we need to add an option to apply activation to each sample before combine operation.

roman-kh: If one needs an activation before combine, one can do it before calling combine.
To include activation in combine, one can add necessary args (layout, activation, etc)

PR #400, opened 2019-09-05
Encoder_decoder improvements

SergeyTsimfer:
This PR proposes to:

- make order of operations configurable in both `encoder` and `decoder`: you can either process tensor via chosen `block` and then change its spatial resolution with `downsampling`/`upsampling`, or vice versa

- `skip` parameter of `decoder/blocks` can now be a dictionary of parameters, that are passed to `combine` function

- now `add_as_function` decorator is smart enough to pass `training` tensor only when needed

### DIFF
--- a/batchflow/models/tf/encoder_decoder.py
+++ b/batchflow/models/tf/encoder_decoder.py
@@ -25,6 +25,9 @@ class EncoderDecoder(TFModel):
             num_stages : int
                 Number of downsampling stages.
 
+            order : sequence
+                Determines order of applying layers: block->downsampling or downsampling->block.
+
             downsample : dict, optional
                 Parameters for downsampling (see :func:`~.layers.conv_block`)
 
@@ -57,6 +60,10 @@ class EncoderDecoder(TFModel):
 
             skip : bool
                 Whether to concatenate upsampled tensor with stored pre-downsample encoding.
+
+            order : sequence
+                Determines order of applying layers: upsampling->block or block->upsampling.
+
             upsample : dict
                 Parameters for upsampling (see :func:`~.layers.upsample`).
 
@@ -120,13 +127,15 @@ class EncoderDecoder(TFModel):
     def default_config(cls):
         config = TFModel.default_config()
 
-        config['body/encoder'] = dict(base=None, num_stages=None, reverse_order=False)
+        config['body/encoder'] = dict(base=None, num_stages=None,
+                                      order=['block', 'downsampling'])
         config['body/encoder/downsample'] = dict(layout='p', pool_size=2, pool_strides=2)
         config['body/encoder/blocks'] = dict(base=cls.block)
 
         config['body/embedding'] = dict(base=cls.block)
 
-        config['body/decoder'] = dict(skip=True, num_stages=None, factor=None, reverse_order=False)
+        config['body/decoder'] = dict(skip=True, num_stages=None, factor=None,
+                                      order=['upsampling', 'block'])
         config['body/decoder/upsample'] = dict(layout='tna')
         config['body/decoder/blocks'] = dict(base=cls.block, combine_op='concat')
         return config
@@ -206,6 +215,9 @@ class EncoderDecoder(TFModel):
         num_stages : int
             Number of downsampling stages.
 
+        order : sequence
+            Determines order of applying layers: block->downsampling or downsampling->block.
+
         blocks : dict
             Parameters for tensor processing before downsampling.
 
@@ -220,9 +232,8 @@ class EncoderDecoder(TFModel):
         list of tf.Tensors
         """
         base_class = kwargs.pop('base')
-        steps, downsample, block_args = cls.pop(['num_stages', 'downsample', 'blocks'], kwargs)
-        reverse_order = kwargs.pop('reverse_order')
-        print('REVERSE ORDER', reverse_order)
+        steps, downsample, block_args, order = cls.pop(['num_stages', 'downsample', 'blocks', 'order'], kwargs)
+        order = ''.join([item[0] for item in order])
 
         if base_class is not None:
             encoder_outputs = base_class.make_encoder(inputs, name=name, **kwargs)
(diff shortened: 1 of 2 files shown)


batchflow/models/tf/encoder_decoder.py
```diff
@@ -25,6 +25,9 @@ class EncoderDecoder(TFModel):
             num_stages : int
                 Number of downsampling stages.
 
+            order : sequence
+                Determines order of applying layers: block->downsampling or downsampling->block.
+
```
a-arefina: please add possible values

batchflow/models/tf/encoder_decoder.py
```diff
@@ -55,8 +58,14 @@ class EncoderDecoder(TFModel):
...
 
-            skip : bool
-                Whether to concatenate upsampled tensor with stored pre-downsample encoding.
+            skip : bool, dict
+                If bool, then whether to combine upsampled tensor with stored pre-downsample encoding.
+                If dict, then parameters for combining upsampled tensor with stored pre-downsample encoding,
+                see :class:`~.layers.Combine`.
+
+            order : sequence
+                Determines order of applying layers: upsampling->block or block->upsampling.
```
a-arefina: double description

batchflow/models/tf/encoder_decoder.py
```diff
@@ -220,7 +244,8 @@ def encoder(cls, inputs, name='encoder', **kwargs):
         list of tf.Tensors
         """
         base_class = kwargs.pop('base')
-        steps, downsample, block_args = cls.pop(['num_stages', 'downsample', 'blocks'], kwargs)
+        steps, order, downsample, block_args = cls.pop(['num_stages', 'order', 'downsample', 'blocks'], kwargs)
+        order = ''.join([item[0] for item in order])
```
a-arefina: what about using this in config directly? order: 'ub'

batchflow/models/tf/encoder_decoder.py
```diff
@@ -233,14 +258,20 @@ def encoder(cls, inputs, name='encoder', **kwargs):
...
-
-                        # Downsampling
-                        if downsample.get('layout') is not None:
-                            args = {**kwargs, **downsample, **unpack_args(downsample, i, steps)}
-                            x = conv_block(x, name='downsample-{}'.format(i), **args)
+                        # Make all the args
+                        args = {**kwargs, **block_args, **unpack_args(block_args, i, steps)}
+                        downsample_args = {**kwargs, **downsample, **unpack_args(downsample, i, steps)}
+
+                        if order in ['bd', 'bp']: # block -> downsample
```
a-arefina: what is `'p'`?
what about checking `order[0] == 'b'`?

batchflow/models/tf/encoder_decoder.py
```diff
@@ -233,14 +258,20 @@ def encoder(cls, inputs, name='encoder', **kwargs):
...
+                            x = base_block(x, name='block', **args)
+                            if downsample.get('layout') is not None:
+                                x = conv_block(x, name='downsample', **downsample_args)
+                        elif order in ['db', 'pb']: # downsample -> block
+                            if downsample.get('layout') is not None:
+                                x = conv_block(x, name='downsample', **downsample_args)
+                            x = base_block(x, name='block', **args)
+                        else:
+                            raise ValueError('Unknown order {}, use one of ("bd", "db")'.format(order))
```
a-arefina: I would check order format and then
```
for comp in order:
    if comp is 'b':
        do block
    else:
       do downsample
```
to avoid code copy-paste

batchflow/models/tf/encoder_decoder.py
```diff
@@ -291,15 +322,31 @@ def decoder(cls, inputs, name='decoder', **kwargs):
...
+            other args : dict
+                Parameters for the base block.
+
         kwargs : dict
             Parameters for ``upsample`` method.
 
+        Notes
+        -----
+        Inputs must be a sequence of encodings, where the last item (inputs[-1]) and
+        the second last (inputs[-2]) have the same spatial shape and thus not used as skip-connections.
```
a-arefina:
```suggestion
        the second last (`inputs[-2]`) have the same spatial shape and thus are not used as skip-connections.
```

batchflow/models/tf/encoder_decoder.py
```diff
@@ -311,7 +358,8 @@ def decoder(cls, inputs, name='decoder', **kwargs):
         """
         steps = kwargs.pop('num_stages') or len(inputs)-2
         factor = kwargs.pop('factor') or [2]*steps
-        skip, upsample, block_args = cls.pop(['skip', 'upsample', 'blocks'], kwargs)
+        skip, order, upsample, block_args = cls.pop(['skip', 'order', 'upsample', 'blocks'], kwargs)
+        order = ''.join([item[0] for item in order])
```
a-arefina: same as in encoder

batchflow/models/tf/encoder_decoder.py
```diff
@@ -328,26 +376,30 @@ def decoder(cls, inputs, name='decoder', **kwargs):
...
+                        x = base_block(x, name='block-', **args)
+                    elif order == 'bu': # block -> upsample
+                        x = base_block(x, name='block', **args)
+                        if upsample.get('layout') is not None:
+                            x = cls.upsample(x, name='upsample', **upsample_args)
+                    else:
+                        raise ValueError('Unknown order {}, use one of ("ub", "bu")'.format(order))
+
+                    # Combine result with the stored encoding of the ~same shape
                     if skip and (i < len(inputs)-2):
```
a-arefina: If skip in config is an empty dict, this check will not pass. Is this what you want?

batchflow/models/tf/encoder_decoder.py
SergeyTsimfer: You can use it, as well as list with full names of operations

batchflow/models/tf/encoder_decoder.py
SergeyTsimfer: `p` stands for `pooling`

batchflow/models/tf/encoder_decoder.py
SergeyTsimfer: That is definitely better. That would also allow to use multiple `blocks` in a row, though I don't know whether we want it.

batchflow/models/tf/encoder_decoder.py
SergeyTsimfer: On one hand, `{}` evaluates to `False` and should be a sign of that.
On the other hand, in this situation it can also mean 'I want to combine tensors with no additional parameters'.

batchflow/models/tf/encoder_decoder.py
```diff
@@ -55,8 +58,14 @@ class EncoderDecoder(TFModel):
                 If int, the total upsampling factor for all stages combined.
                 If list, upsampling factors for each stage.
 
-            skip : bool
-                Whether to concatenate upsampled tensor with stored pre-downsample encoding.
+            skip : bool, dict
+                If bool, then whether to combine upsampled tensor with stored pre-downsample encoding.
```
roman-kh: specify which combine operation is used

batchflow/models/tf/encoder_decoder.py
```diff
@@ -65,6 +74,8 @@ class EncoderDecoder(TFModel):
 
                 base : callable
                     Tensor processing function. Default is :func:`~.layers.conv_block`.
+                combine_op : str
```
roman-kh: dict?

PR #401, opened 2019-09-05
Cyclical learning rate 

Dimonovez:
This PR proposes to add cyclical learning rate.


batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,74 @@
...
+    """Applies cyclic learning rate (CLR).
+    https://arxiv.org/pdf/1506.01186.pdf.
+
+    This function varies the learning rate between the
+    minimum (learning_rate) and the maximum (max_lr).
+    It returns the cyclic learning rate.
+
+    Args:
+    learning_rate: A scalar `float32` or `float64` `Tensor` or a
+        Python number.  The initial learning rate which is the lower bound
```
a-arefina: here you write it's initial learning rate, but in `mode` above there are different initial values/ please clarify

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,74 @@
...
+    mode:
+        If 'sin' or 'sine' or 'sine wave':
+            Learning rate changes as a sine wave, starting
+            from (max_lr-learning_rate)/2 then decreasing to `learning_rate`.
+        If 'triangular' or 'triangular wave' or 'zigzag':
+            Default, linearly increasing then linearly decreasing the
+            learning rate at each cycle. Learning rate starting
+            from (max_lr-learning_rate)/2 then decreasing to `learning_rate`.
+        If 'sawtooth' or 'saw' or 'sawtooth wave' or 'saw wave':
+            Linearly increasing to `max_lr` then again increasing from `learning_rate`
```
a-arefina: how this differs from 'trianguar'?

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,74 @@
...
+            learning rate at each cycle. Learning rate starting from `learning_rate`
+            then increasing.
+    name: String.  Optional name of the operation.  Defaults to
+        'CyclicLearningRate'.
+
+    Returns:
+    A scalar `Tensor` of the same type as `learning_rate`. The cyclic
+    learning rate.
+    """
+    with ops.name_scope_v2(name or "CyclicLearningRate"):
```
a-arefina: why not pass "CyclicLearningRate" as default value directly?

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,74 @@
...
+        # ((max_lr - learning_rate)/pi) * asin(sin(((2 * -pi)/step_size) * global_step)) + (max_lr + learning_rate)/2
+            first_factor = math_ops.divide(math_ops.subtract(max_lr, learning_rate), pi)
+            inside_sin = math_ops.multiply(math_ops.divide(math_ops.multiply(2., -pi), step_size), global_step)
+            second_factor = asin(sin(inside_sin))
+        elif mode in ('sawtooth', 'saw', 'sawtooth wave', 'saw wave'):
+        # (max_lr - learning_rate) * (global_step/step_size - floor(global_step/step_size)) + learning_rate
+            first_factor = math_ops.subtract(max_lr, learning_rate)
+            divided_global_step = math_ops.divide(global_step, step_size)
+            second_factor = math_ops.subtract(divided_global_step, floor(divided_global_step))
+            return math_ops.add(math_ops.multiply(first_factor, second_factor), learning_rate)
```
a-arefina: I would add `second_comp` inside each branch and use single return statement

batchflow/models/tf/nn/train.py
Dimonovez: In `saw` mode learning rate linearly increases from `learning rate` to `max_lr` and then sharply drops to `learning rate`.

batchflow/models/tf/nn/train.py
a-arefina: Well, I actually meant adding single return in functions bottom outside if-else branches 
`return math_ops.add(math_ops.multiply(first_factor, second_factor), second_comp)`

batchflow/models/tf/nn/train.py
a-arefina: Add this to docs

batchflow/models/tf/nn/train.py
Dimonovez: It's tensorflow-like signature of function with optional `name` of scope.

Dimonovez: We use `name_scope` when we don't want to interfere on that scope. If `name=None` scope would named `CyclicLearningRate` and and wouldn't reset scope to the top level.

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,95 @@
...
+            ```
+
+    name: str
+        Optional name of the operation (default='CyclicLearningRate')
+
+    Returns
+    -------
+    tf.Tensor
+    """
+    with tf.name_scope(name or "CyclicLearningRate"):
```
roman-kh: add a default value to `name` arg

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,95 @@
...
+        learning_rate = tf.cast(learning_rate, dtype=tf.float32)
+        global_step = tf.cast(global_step, dtype=tf.float32)
+        step_size = tf.cast(step_size, dtype=tf.float32)
+        max_lr = tf.cast(max_lr, dtype=tf.float32)
+
+        if mode in ('sin', 'sine', 'sine wave'):
```
roman-kh: You don't need too many synonyms. Just 'sin', 'tri' and 'saw'.

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def piecewise_constant(global_step, *args, **kwargs):
+    """ Constant learning rate decay (uses global_step param instead of x) """
+    return tf.train.piecewise_constant(global_step, *args, **kwargs)
+
+def cyclic_learning_rate(learning_rate, global_step, max_lr=0.1, step_size=10,
+                         mode='triangular', name='CyclicLearningRate'):
+    """ Applies cyclic learning rate (CLR).
+    https://arxiv.org/abs/1506.01186.
```
dpodvyaznikov: Use common style for mentioning publications. See `TFModel.se_block` for example.

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,97 @@
...
+    minimum (learning_rate) and the maximum (max_lr).
+    It returns the decayed learning rate.
+
+    Leslie N. Smith "`Cyclical Learning Rates
+    for Training Neural Networks <https://arxiv.org/abs/1506.01186>`_"
+
+    Parameters
+    ----------
+    learning_rate: float or tf.Tensor
+        the minimum learning rate boundary
```
SergeyTsimfer: Use this [guide](https://numpydoc.readthedocs.io/en/latest/format.html) in the docstring

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,97 @@
...
+    global_step: int or tf.Tensor
+        global step to use for the cyclic computation. Must not be negative
+    max_lr: float
+        the maximum learning rate boundary (default=0.1)
+    step_size: int
+        the number of iterations in half a cycle (default=10)
+    mode:
```
SergeyTsimfer: Add description of the parameter itself, not only the possible values

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,97 @@
+""" Helpers for training """
+from math import pi
+
+import tensorflow as tf
+from tensorflow.math import sin, asin, floor # pylint: disable=import-error
```
SergeyTsimfer: Each of these functions if used pretty much only once, so it would be better to call them as `tf.math.<>`

dpodvyaznikov: Why?

batchflow/models/tf/nn/train.py
Dimonovez: Should I use the  'References' section to refer to the article and the 'Notes' section to show how functions are computed?

batchflow/models/tf/nn/train.py
Dimonovez: @SergeyTsimfer, Done

roman-kh:
Refrain from naming your commits as "Fix review". Declare what exactly has changed

batchflow/models/tf/nn/train.py
SergeyTsimfer: > Why?

The next person to add smth to this file will need to either add more function (and thus clutter the beginning of the file), or change usage to directly to `tf`, as I suggested

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Leslie N. Smith "`Cyclical Learning Rates
+    for Training Neural Networks <https://arxiv.org/abs/1506.01186>`_"
+
+    Parameters
+    ----------
+    learning_rate : float or tf.Tensor
+        The minimum learning rate boundary.
+    global_step: int or tf.Tensor
+        Global step to use for the cyclic computation. Must not be negative.
```
SergeyTsimfer: Unclear

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,97 @@
...
+    max_lr : float
+        The maximum learning rate boundary.
+    step_size : int, optional
```
SergeyTsimfer: It is not optional: it has default value

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,97 @@
...
+        The number of iterations in half a cycle (the default is 10).
+    mode : {'tri', 'sin', 'saw'}, opional
```
SergeyTsimfer: Same

batchflow/models/tf/nn/train.py
```diff
@@ -0,0 +1,97 @@
...
+        Set the learning rate change function.
```
roman-kh: Move long description of modes to notes section

batchflow/models/tf/nn/train.py
Dimonovez: Why? I think the [guide](https://numpydoc.readthedocs.io/en/latest/format.html) says different.

PR #415, opened 2019-09-17
Efficientnet

a-arefina:
This PR proposes ScalableModel Base  class and batchflow EfficientNet family implementation

Notebooks created in this PR are for debug purposes and will be removed before merging into master


batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+
+        if config.get('head/units') is None:
+            config['head/units'] = self.num_classes('targets')
+
+        return config
+
+    @classmethod
+    def round_filters(cls, filters, factor):
+        """ Round number of filters based on width multiplier. """
+        return int(filters * factor)
```
SergeyTsimfer: Why no `ceil` in this case?

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+
+    @classmethod
+    def round_repeats(cls, repeats, factor):
+        """ Round number of layers based on depth multiplier. """
+        return int(np.ceil(repeats * factor))
+
+    @classmethod
+    def initial_block(cls, inputs, name='initial_block', **kwargs):
```
SergeyTsimfer: Not needed

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+        with tf.variable_scope(name):
+            blocks = kwargs.pop('blocks')
+            x = inputs
+            for i, block_args in enumerate(blocks):
+                x = cls.block(x, name='block-%d' % i, **{**kwargs, **block_args})
+
+            return x
+
+    @classmethod
+    def head(cls, inputs, name='head', **kwargs):
```
SergeyTsimfer: Not needed

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+                x = inputs
+                for i in range(repeats):
+                    x = conv_block(x, name='layer-%d' % i, **new_layer_args)
+
+                return x
+        else:
+            return base_block(inputs, name=name, **new_layer_args)
+
+    @classmethod
+    def update_layer_params(cls, kwargs):
```
SergeyTsimfer: Make so that function and `round_filters`/`round_repeates` are in the same location of the code: that would increase readability

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+
+    @classmethod
+    def round_filters(cls, filters, factor):
+        """ Round number of filters based on width multiplier. """
+        return int(filters * factor)
+
+    @classmethod
+    def round_repeats(cls, repeats, factor):
+        """ Round number of layers based on depth multiplier. """
+        return int(np.ceil(repeats * factor))
```
SergeyTsimfer: Better question: why `ceil` here?

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+        """ calculate filters and number of repeats depending on scaling factors """
+        new_kwargs = dict(**kwargs)
+        scalable = new_kwargs.get('scalable', False)
+
+        if not scalable:
+            return new_kwargs
+
+        w_factor = new_kwargs.get('width_factor')
+        filters = new_kwargs.get('filters')
+        if filters and w_factor > 1:
```
SergeyTsimfer: Should not we allow for lessening the network by checking `factor != 1`?

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+    def update_layer_params(cls, kwargs):
+        """ calculate filters and number of repeats depending on scaling factors """
+        new_kwargs = dict(**kwargs)
+        scalable = new_kwargs.get('scalable', False)
+
+        if not scalable:
+            return new_kwargs
+
+        w_factor = new_kwargs.get('width_factor')
+        filters = new_kwargs.get('filters')
```
SergeyTsimfer: It would not make a difference, but probably the intention was to use `kwargs` to get both `filters` and `w_factor`

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+            elif isinstance(filters, list):
+                new_kwargs['filters'] = [cls.round_filters(f, w_factor) for f in filters]
+            else:
+                raise ValueError("filters should be int or list, {} given".format(type(filters)))
+
+        d_factor = kwargs.get('depth_factor')
+        repeats = kwargs.get('repeats')
+        if repeats and d_factor > 1:
+            new_kwargs['repeats'] = cls.round_repeats(repeats, d_factor)
+
```
SergeyTsimfer: The `repeats` and `filters` are processed in quite similar fashion: you get some `name` from `kwargs` along with needed `factor` and apply `round_name` function to the former (with some type and error handling). The actual difference being only in used `factor`: width or depth. That can be generalized quite well (with the aim on other sizing parameters)

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+    resolution = 224  # image resolution used in original paper
+
+    @classmethod
+    def default_config(cls):
+        config = super().default_config()
+
+        config['initial_block'].update(dict(scalable=True, layout='cna', kernel_size=3, strides=2, filters=32))
+
+        config['body'].update(dict(base=MobileNet_v2.block, blocks=[
```
SergeyTsimfer: Don't you think that dictionary with parameters, each of which is `list`, would look nicer? You would be able to slice them just like `EncoderDecoder` does, and eliminate the need to use list of dictionaries with huge redundancy in name-typing

batchflow/models/tf/efficientnet.py
SergeyTsimfer: Actually, you can bilinearly `resize` the input image to desired resolution right here, with `resolution_factor`  or `resolution` involved.

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+
+"""
+import numpy as np
+import tensorflow as tf
+
+from . import TFModel, MobileNet_v2
+from .layers import conv_block
+
+
+class ScalableModel(TFModel):
```
SergeyTsimfer: Overall, that looks great. My only concern is that you use `factors` to modify different parameters right before their application. Why not just scan the whole `config` once and update all the values?

Just a quick thought: it is possible to make `@scalable` decorator and apply that to any of the existing class. Under the hood, that decorator would overload `build_config` method by:

- calling the `build_config` (that is, from the decorated class)

- applying the scaling logic to the confiuration-dictionary

Obviously, there are some questions (for example, what should be the right place for `factor's` in this case).

batchflow/models/tf/efficientnet.py
SergeyTsimfer: It would be better to use fair `round` before `int`, which always `floors` the value

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,318 @@
...
+                indicates whether the block can be scaled
+            other parameters :
+                see :func:`~.layers.conv_block` or base block documentation
+
+    head : dict
+        parameters for head
+
+    common : dict
+        common parameters
+        width_factor :
```
Dimonovez: typo

batchflow/models/tf/efficientnet.py
a-arefina: I will consider modifying config in build_config directly, but as for now we do need body to iterate over blocks and to process `repeats` properly if we use something other than MobileNet_v2.block

batchflow/models/tf/efficientnet.py
a-arefina: > Better question: why `ceil` here?
This comes directly from original implementation of EfficientNet

perhaps this method should be abstract

batchflow/models/tf/efficientnet.py
a-arefina: Yes, we should

batchflow/models/tf/efficientnet.py
a-arefina: It is needed for EfficientNets which have initial_blocks and heads that are width-scalable and not depth-scalabe. I'll move it there

batchflow/models/tf/efficientnet.py
a-arefina: This won't work for blocks that have different bases

SergeyTsimfer:
<p>Test the same pipeline on <code>MobileNet</code>/<code>ResNet</code> with different dataset (<code>ImageWoof</code>/<code>ImageNette</code>)</p>

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,252 @@
...
+        d_factor = config.get('common/depth_factor')
+
+        for path in ('initial_block/', 'body/encoder/blocks/', 'head/'):
+            scalable = config.get(path + 'scalable')
+            if scalable:
+                for param, factor, fun in [('filters', w_factor, self.round_filters),
+                                           ('repeats', d_factor, self.round_repeats)]:
+                    if factor != 1:
+                        val = config.get(path + param)
+                        if val:
```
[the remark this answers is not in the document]
a-arefina: why?

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,252 @@
...
+        scalable : bool
+            indicates whether the block can be scaled
+
+    common : dict, optional
+        common parameters
+        width_factor, depth_factor : float, optional
+            scaling factors
+    """
+
+    resolution = 224  # image resolution used in original paper
```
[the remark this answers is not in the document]
a-arefina: These values are defined in the original paper.
One could refer to them in crop or resize actions, for example: `EfficientNetB1.resolution`

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,252 @@
...
+
+        return config
+
+    @classmethod
+    def round_filters(cls, filters, factor):
+        """ Round number of filters based on width multiplier. """
+        return max(1, int(filters * factor))
+
+    @classmethod
+    def round_repeats(cls, repeats, factor):
```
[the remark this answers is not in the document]
a-arefina: We would needed them if we decided to  make other scalable models with more complex rounding method

batchflow/models/tf/efficientnet.py
```diff
@@ -0,0 +1,242 @@
...
+        parameters for the initial block (see :class:`~tf.EncoderDecoder`)
+        scalable : bool
+            indicates whether the block can be scaled
+
+    body : dict, optional
+        encoder : dict, optional
+            parameters for model's body (see :class:`~tf.EncoderDecoder`)
+
+            num_stages : int, optional
+                Number of blocks
```
SergeyTsimfer: Some of the docstrings are formatted in such a way that every description starts with a capital letter and ends with a point, some of them don't. Choose one:)

batchflow/models/tf/efficientnet.py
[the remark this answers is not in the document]
SergeyTsimfer: In most cases `factor` is not equal to one, so it would result in less checks on average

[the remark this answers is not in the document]
a-arefina: Really not sure about "most cases" and what the average checks number would be... It will depend on use cases. This piece of code is definitely not a bottleneck so I would prefer to leave it as is

PR #416, opened 2019-09-19
Unet build config

a-arefina:
build_config for Unet to make default parameters
+ test fix

### DIFF
--- a/batchflow/models/tf/encoder_decoder.py
+++ b/batchflow/models/tf/encoder_decoder.py
@@ -27,10 +27,10 @@ class EncoderDecoder(TFModel):
 
             order : str, sequence of str
                 Determines order of applying layers.
-                If str, then each letter stands for operation: 'b' for block, 'd'/'p' for downsampling.
+                If str, then each letter stands for operation:
+                'b' for 'block', 'd'/'p' for 'downsampling', 's' for 'skip'.
                 If sequence, than the first letter of each item stands for operation:
-                'b' for block, 'd' for downsampling.
-                For example, `'bd'` allows to use block->downsampling.
+                For example, `'sbd'` allows to use throw skip connection -> block -> downsampling.
 
             downsample : dict, optional
                 Parameters for downsampling (see :func:`~.layers.conv_block`)
@@ -70,9 +70,10 @@ class EncoderDecoder(TFModel):
 
             order : str, sequence of str
                 Determines order of applying layers.
-                If str, then each letter stands for operation: 'b' for block, 'u' for upsampling.
-                If sequence, than the first letter of each item stands for operation: 'b' for block, 'u' for upsampling.
-                For example, `'ub'` allows to use upsampling->block.
+                If str, then each letter stands for operation:
+                'b' for 'block', 'u' for 'upsampling', 'c' for 'combine'
+                If sequence, than the first letter of each item stands for operation.
+                For example, `'ucb'` allows to use upsampling-> combine ->block.
 
             upsample : dict
                 Parameters for upsampling (see :func:`~.layers.upsample`).
@@ -141,14 +142,14 @@ class EncoderDecoder(TFModel):
         config = TFModel.default_config()
 
         config['body/encoder'] = dict(base=None, num_stages=None,
-                                      order=['block', 'downsampling'])
+                                      order=['skip', 'block', 'downsampling'])
         config['body/encoder/downsample'] = dict(layout='p', pool_size=2, pool_strides=2)
         config['body/encoder/blocks'] = dict(base=cls.block)
 
         config['body/embedding'] = dict(base=cls.block)
 
         config['body/decoder'] = dict(skip=True, num_stages=None, factor=None,
-                                      order=['upsampling', 'block', 'concat'])
+                                      order=['upsampling', 'block', 'combine'])
         config['body/decoder/upsample'] = dict(layout='tna')
         config['body/decoder/blocks'] = dict(base=cls.block, combine_op='concat')
         return config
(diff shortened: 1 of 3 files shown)


batchflow/models/tf/encoder_decoder.py
```diff
@@ -148,7 +148,7 @@ def default_config(cls):
         config['body/embedding'] = dict(base=cls.block)
 
         config['body/decoder'] = dict(skip=True, num_stages=None, factor=None,
-                                      order=['upsampling', 'block'])
+                                      order=['upsampling', 'block', 'concat'])
```
SergeyTsimfer: Rename it to `combine`: we can use not only `concat` as `combine_op`

batchflow/models/tf/encoder_decoder.py
```diff
@@ -411,13 +411,14 @@ def decoder(cls, inputs, name='decoder', **kwargs):
                         elif letter in ['u']:
                             if upsample.get('layout') is not None:
                                 x = cls.upsample(x, name='upsample', **upsample_args)
+                        elif letter == 'c':
```
SergeyTsimfer: Maybe add `s` (stands for `skip`)?

batchflow/models/tf/unet.py
```diff
@@ -59,14 +59,35 @@ class UNet(EncoderDecoder):
     def default_config(cls):
         config = super().default_config()
 
-        config['initial_block'] = dict(layout='cna cna', kernel_size=3, filters=64)
+        config['initial_block'] = dict(layout='cna cna', kernel_size=3)
 
         config['body/encoder/num_stages'] = 4
-        config['body/encoder/blocks'] = dict(layout='cna cna', kernel_size=3, filters=[128, 256, 512, 1024])
+        config['body/encoder/order'] = ['downsampling', 'block']
+        config['body/encoder/blocks'].update(dict(layout='cna cna', kernel_size=3))
```
SergeyTsimfer: Why `update`, not `assign`? It should work just fine (I double checked)

batchflow/models/tf/unet.py
```diff
@@ -59,14 +59,35 @@ class UNet(EncoderDecoder):
...
+        num_stages = config.get('body/encoder/num_stages')
+
+        if config.get('body/encoder/blocks/filters') is None:
+            config['body/encoder/blocks/filters'] = [ibf * 2 * 2**i for i in range(num_stages)]
+
+        if config.get('body/decoder/blocks/filters') is None:
+            enc_filters = config.get('body/encoder/blocks/filters')
+            config['body/decoder/blocks/filters'] = [f for f in reversed(enc_filters[:-1])] + [ibf]
+
+        if config.get('body/decoder/upsample/filters') is None:
```
SergeyTsimfer: Not needed

batchflow/models/tf/unet.py
```diff
@@ -59,14 +59,35 @@ class UNet(EncoderDecoder):
...
+        config['body/decoder/order'] = ['upsampling', 'concat', 'block']
+        config['body/decoder/blocks'].update(dict(layout='cna cna', kernel_size=3))
 
         config['loss'] = 'ce'
         return config
+
+    def build_config(self, names=None):
+        config = super().build_config(names)
+
+        ibf = config.get('initial_block/filters')
```
SergeyTsimfer: Do we need `initial_block` at all?
All the changes to `EncoderDecoder` were supposed to make it so we don't need it, and use `embedding` for the widest part (1024 filters) of the network.
This way there will be no shift in number of filters in `encoder`/`decoder`

batchflow/models/tf/encoder_decoder.py
a-arefina: I think it will be misleading. We are combining inputs here, not adding skips.

batchflow/models/tf/unet.py
a-arefina: In Unet upsample block both increases resolution and reduces width, so we have to set upsample filters. If we don't, upsample operation doesn't modify the number of channels

batchflow/models/tf/unet.py
a-arefina: assignment works similarly as update indeed, but this is really misleading

made assignment here for now, but created issue https://github.com/analysiscenter/batchflow/issues/417 after discussion with @roman-kh

batchflow/models/tf/unet.py
```diff
@@ -59,14 +57,32 @@ class UNet(EncoderDecoder):
...
         config['loss'] = 'ce'
         return config
+
+    def build_config(self, names=None):
+        config = super().build_config(names)
+
+        num_stages = config.get('body/encoder/num_stages')
+
+        if config.get('body/encoder/blocks/filters') is None:
+            config['body/encoder/blocks/filters'] = [64 * 2**i for i in range(num_stages)]
```
a-arefina: If we want to control the number of filters in 1st encoder bock, this can be easily done. The main problem with it is how to name this parameter and where to place it in config. Any suggestions are welcome!

batchflow/models/tf/unet.py
```diff
@@ -59,14 +57,32 @@ class UNet(EncoderDecoder):
     def default_config(cls):
```
SergeyTsimfer: We need default values for filters here, so one could use `body` of this class in custom classes.

a-arefina: There are other models where filters are not set in default_config, for example ResNet. I see 2 options: 
1) if one decides to use some class's body, they have to provide all needed parameters explicitly
2) call `build_config` in `fill_params`

batchflow/models/tf/unet.py
SergeyTsimfer: Maybe, `config['common/filters_base']`?

PR #424, opened 2019-10-03
Add more exception types

SergeyTsimfer:
`os.path.split` raises different type of excpetions on python 3.7 and 3.5.

image


dpodvyaznikov:
I'd suggest to add a check that `path` is a string in the beginning of `build_from_one_path`, and throw `TypeError` otherwise.
This way this test can be either removed, or modified to catch `TypeError` only.

PR #420, opened 2019-09-26
Fix docs

SergeyTsimfer:
This PR:

- Adds and improves a lot of docstrings in `TFModel` and `TorchModel`

- Fixes warnings during docs autogeneration by improving docs of `ImagesBatch`, `Pipeline`, `Research` and `NamedExpressions`. See #414 

- Changes the style of doc to `sphinx_rtd_theme`

PR #425, opened 2019-10-03
Change loss and decay

SergeyTsimfer:
Contribution of this PR is two-fold:

- Make `loss` capable of working with functions that use more than two tensors to compute value of loss

- Slightly refactor `_make_loss` and `make_decay`

Note that this PR needs #420 to be merged beforehand to add documentation in better format.


roman-kh:
What about `torch`?

PR #427, opened 2019-10-08
Fix P expression

AlexeyKozhevin:
Change `name` evaluation logic to allow constructions such as
`origin = P((R('uniform'), R('uniform')))`


roman-kh:
Pull `master` to get new checks

PR #433, opened 2019-10-21
Named expr

AlexeyKozhevin:
Change named expressions to make the class more flexible and to inherit ResearchNamedExpressions from it. 

- `eval_expr` became class method
- parameters of named expressions are defined as class parameter

pylint check failed because of non changed files.

### DIFF
--- a/batchflow/named_expr.py
+++ b/batchflow/named_expr.py
@@ -15,6 +15,8 @@ def swap(op):
         return op(b, a)
     return _op_
 
+def eval_expr(expr, *args, **kwargs):
+    return NamedExpression.eval_expr(expr, *args, **kwargs)
 
 AN_EXPR = "#!__op__"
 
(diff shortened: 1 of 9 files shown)


batchflow/named_expr.py
```diff
@@ -4,46 +4,11 @@
...
 import numpy as np
 
-
 class _DummyBatch:
     """ A fake batch for static models """
     def __init__(self, pipeline):
         self.pipeline = pipeline
 
-
-def eval_expr(expr, batch=None, pipeline=None, model=None):
```
roman-kh: There is no need to remove `eval_expr`. It can take a named expr (or a bunch of exprs) and evaluate it (them) against kwargs given. So you will use `eval_expr` instead a much longer `NamedExpression.eval_expr`

AlexeyKozhevin: I've returned it.

PR #434, opened 2019-10-21
PSPNet

dpodvyaznikov:
PR proposes following changes:
* introduce a PSPNet model;
* add `body/donwsample` parameter for ResNet models for better control of spatial dimensions: it should be a list of lists with same lenght as `body/num_blocks`, so that each list represents a group. Inner lists contain integers - indices of ResNet blocks that should apply downsampling; empty list indicates that there is no downsampling in group;
* change config parsing in `EncoderDecoder.encoder` method. Now, in case `body/encoder/base` is present, it is built with all kwargs except for `body/encoder/order`. This allowed for control of downsampling scheme in case of ResNet backbone;
* modify tests according to changes decribed above; add test case for PSPNet compilation;
* update pylint exceptions;


roman-kh:
Please, also add a long-time anticipated change: get rid of unneeded params (`resnext_factor` and `bottleneck_factor`), there is `resnext` and `bottleneck` which fit the purpose quite well.

roman-kh:
And reconcile `torch.ResNet` as much as possible

PR #451, opened 2019-11-21
Unet++

AlexeyKozhevin:
- Add UNet++ model

- Add multiple outputs to body

### DIFF
--- a/batchflow/models/tf/base.py
+++ b/batchflow/models/tf/base.py
@@ -1587,7 +1587,7 @@ class TFModel(BaseModel):
             return conv_block(inputs, name=name, **kwargs)
         return inputs
 
-    def output(self, inputs, predictions=None, ops=None, prefix=None, **kwargs):
+    def output(self, inputs, predictions=None, ops=None, **kwargs):
         """ Add output operations to the model graph, like predicted probabilities or labels, etc.
 
         Parameters
@@ -1667,7 +1667,7 @@ class TFModel(BaseModel):
                     self._add_output_op(tensor, oper, oper, attr_prefix, attr_postfix, **kwargs)
 
                 if ctx:
-                    ctx.__exit__(None, None, None) 
+                    ctx.__exit__(None, None, None)
 
         _predictions = [self.get_from_attr('predictions_' + str(i)) for i in range(len(inputs))]
         self._add_predictions(_predictions, 'predictions', '', '', **kwargs)
--- a/batchflow/models/tf/base.py
+++ b/batchflow/models/tf/base.py
@@ -1726,7 +1726,7 @@ class TFModel(BaseModel):
         if len(predictions) > 1:
             x = tf.stack(predictions, name=name)
         else:
-            x = tf.identity(predictions, name=name)
+            x = tf.identity(predictions[0], name=name)
         self.store_to_attr(attr_prefix + name + attr_postfix, x, device)
 
     @classmethod
--- a/batchflow/models/tf/base.py
+++ b/batchflow/models/tf/base.py
@@ -1648,73 +1648,86 @@ class TFModel(BaseModel):
 
         if not isinstance(inputs, (tuple, list)):
             inputs = [inputs]
-
         for i, tensor in enumerate(inputs):
             if not isinstance(tensor, tf.Tensor):
                 raise TypeError("Network output is expected to be a Tensor, but given {}".format(type(inputs)))
 
-            prefix = [*ops.keys()][i]
-            if prefix:
-                ctx = tf.variable_scope(prefix)
-                ctx.__enter__()
-            else:
-                ctx = None
-            attr_prefix = prefix + '_' if prefix else ''
+            self._add_output_op(tensor, predictions, 'predictions', '', '_' + str(i), **kwargs)
 
-            self._add_output_op(tensor, predictions, 'predictions', '', **kwargs)
-            for oper in ops[prefix]:
-                self._add_output_op(tensor, oper, oper, attr_prefix, **kwargs)
+            for prefix in ops.keys():
+                if prefix:
+                    ctx = tf.variable_scope(prefix)
+                    ctx.__enter__()
+                else:
+                    ctx = None
+                attr_prefix = prefix + '_' if prefix else ''
+                attr_postfix = '_' + str(i) if len(inputs) > 1 else ''
 
-            if ctx:
-                ctx.__exit__(None, None, None)
+                for oper in ops[prefix]:
+                    self._add_output_op(tensor, oper, oper, attr_prefix, attr_postfix, **kwargs)
 
-    def _add_output_op(self, inputs, oper, name, attr_prefix, **kwargs):
+                if ctx:
+                    ctx.__exit__(None, None, None) 
+
+        _predictions = [self.get_from_attr('predictions_' + str(i)) for i in range(len(inputs))]
+        self._add_predictions(_predictions, 'predictions', '', '', **kwargs)
+
+    def _add_output_op(self, inputs, oper, name, attr_prefix, attr_postfix, **kwargs):
         device = self._get_current_device()
         if oper is None:
-            self._add_output_identity(inputs, name, attr_prefix[:-1], device, **kwargs)
+            self._add_output_identity(inputs, name, attr_prefix[:-1], attr_postfix, device, **kwargs)
         elif oper == 'softplus':
-            self._add_output_softplus(inputs, name, attr_prefix, device, **kwargs)
+            self._add_output_softplus(inputs, name, attr_prefix, attr_postfix, device, **kwargs)
         elif oper == 'sigmoid':
-            self._add_output_sigmoid(inputs, name, attr_prefix, device, **kwargs)
+            self._add_output_sigmoid(inputs, name, attr_prefix, attr_postfix, device, **kwargs)
         elif oper == 'proba':
-            self._add_output_proba(inputs, name, attr_prefix, device, **kwargs)
+            self._add_output_proba(inputs, name, attr_prefix, attr_postfix, device, **kwargs)
         elif oper == 'labels':
-            self._add_output_labels(inputs, name, attr_prefix, device, **kwargs)
+            self._add_output_labels(inputs, name, attr_prefix, attr_postfix, device, **kwargs)
         elif callable(oper):
-            self._add_output_callable(inputs, oper, None, attr_prefix, device, **kwargs)
+            self._add_output_callable(inputs, oper, None, attr_prefix, attr_postfix, device, **kwargs)
 
-    def _add_output_identity(self, inputs, name, attr_prefix, device, **kwargs):
+    def _add_output_identity(self, inputs, name, attr_prefix, attr_postfix, device, **kwargs):
         _ = kwargs
         x = tf.identity(inputs, name=name)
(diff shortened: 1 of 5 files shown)


batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+
+        x = super().decoder(inputs, name=name, filters=decoder_filters,
+                            return_all=False, **{**kwargs, 'num_stages': num_stages})
+
+        semantic_outputs.append(x)
+
+        return [semantic_outputs[i] for i in n_outputs]
+
+    @classmethod
+    def head(cls, inputs, targets, name='head', **kwargs):
```
roman-kh: Document somewhere that `head` always expects a sequence of inputs

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+
+        semantic_outputs = []
+        n_outputs = kwargs['n_outputs']
+        if isinstance(n_outputs, int):
+            n_outputs = list(range(-n_outputs, 0))
+
+        for i in range(1, num_stages):
+            _inputs = inputs[:i] + [inputs[i]] * 2
+
+            _kwargs = kwargs
```
roman-kh: Not needed

batchflow/models/tf/unet.py
```diff
@@ -58,14 +66,7 @@ class UNet(EncoderDecoder):
...
 
-        config['body/encoder/num_stages'] = 4
-        config['body/encoder/order'] = ['block', 'skip', 'downsampling']
-        config['body/encoder/blocks'] += dict(layout='cna cna', kernel_size=3, filters=[64, 128, 256, 512])
-        config['body/embedding'] += dict(layout='cna cna', kernel_size=3, filters=1024)
-        config['body/decoder/order'] = ['upsampling', 'combine', 'block']
-        config['body/decoder/blocks'] += dict(layout='cna cna', kernel_size=3, filters=[512, 256, 128, 64])
-
-        config['loss'] = 'ce'
+        config['body/n_outputs'] = 1
```
SergeyTsimfer: Why did you delete most of the default config?

batchflow/models/tf/base.py
```diff
@@ -1587,13 +1587,13 @@ def head(cls, inputs, name='head', **kwargs):
             return conv_block(inputs, name=name, **kwargs)
         return inputs
 
-    def output(self, inputs, predictions=None, ops=None, prefix=None, **kwargs):
+    def output(self, inputs, predictions=None, ops=None, **kwargs):
```
SergeyTsimfer: Can you elaborate on what is going on here besides some code refactoring (better name making for added output `ops` and removing redundant `prefix` argument)?

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+            config['body/decoder/dense/kernel'] = 3
+
+        if config.get('body/decoder/dense/combine_op') is None:
+            config['body/decoder/dense/combine_op'] = 'concat'
+
+        return config
+
+    @classmethod
+    def decoder(cls, inputs, name='decoder', **kwargs):
+        decoder_filters = kwargs['upsample']['filters']
```
SergeyTsimfer:
```suggestion
        decoder_filters = kwargs['upsample/filters']
```
can improve readability everywhere

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+                        'data_format': kwargs.get('data_format'),
+                        **(combine_op if isinstance(combine_op, dict) else {})}
+
+        semantic_outputs = []
+        n_outputs = kwargs['n_outputs']
+        if isinstance(n_outputs, int):
+            n_outputs = list(range(-n_outputs, 0))
+
+        for i in range(1, num_stages):
+            _inputs = inputs[:i] + [inputs[i]] * 2
```
SergeyTsimfer: `_inputs` is used only once, so you don't really need it here

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+
+            _kwargs = kwargs
+            _kwargs['upsample'] = _kwargs['dense']['upsample']
+            _kwargs['upsample']['filters'] = decoder_filters[-i:]
+            _kwargs['blocks']['filters'] = decoder_filters[-i:]
+
+            outputs = super().decoder(_inputs, name=name,
```
SergeyTsimfer: Same `name` would create scopes like `decoder`, `decoder_1`, `decoder_2` etc, while it can be better to reserve name `decoder` for the deepest branch

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+
+                dense_args = {**unpack_args(kwargs['dense'], j, num_stages), **kwargs}
+                x = conv_block(x, name='x-{}-{}'.format(j, i), **dense_args)
+
+                if j == 0:
+                    semantic_outputs.append(x)
+
+                inputs[j] = combine([x, inputs[j]], name='concat-{}-{}'.format(j, i-j), **combine_args)
+
+        kwargs['upsample']['filters'] = decoder_filters
```
SergeyTsimfer: Do you really need it? You don't seem to alter `kwargs` in any way

batchflow/models/tf/unet.py
SergeyTsimfer: Make a proper copy

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+            _kwargs = kwargs
+            _kwargs['upsample'] = _kwargs['dense']['upsample']
+            _kwargs['upsample']['filters'] = decoder_filters[-i:]
+            _kwargs['blocks']['filters'] = decoder_filters[-i:]
+
+            outputs = super().decoder(_inputs, name=name,
+                                      return_all=True, **{**_kwargs, 'num_stages': i})
+
+            for j, x in enumerate(inputs[:i]):
+                x = combine([x, outputs[::-1][j]], **combine_args)
```
SergeyTsimfer:
```suggestion
                x = combine([x, outputs[-j]], **combine_args)
```
that would be easier to read. Or you could use `zip` over `reversed` as iterator

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+
+                dense_args = {**unpack_args(kwargs['dense'], j, num_stages), **kwargs}
```
SergeyTsimfer:
```suggestion
                dense_args = {**kwargs, **unpack_args(kwargs['dense'], j, num_stages)}
```
Current version seems to be working only because `conv_block` applies essentially the same `unpack_args` under the hood. You can even remove this `unpacking`, but explicit is better that implicit, right?

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+                x = conv_block(x, name='x-{}-{}'.format(j, i), **dense_args)
+
+                if j == 0:
+                    semantic_outputs.append(x)
+
+                inputs[j] = combine([x, inputs[j]], name='concat-{}-{}'.format(j, i-j), **combine_args)
```
SergeyTsimfer: Can't find that in the original paper, though it does not seem to be a bad idea

batchflow/models/tf/unet.py
```diff
@@ -87,3 +88,140 @@ def build_config(self, names=None):
...
+        semantic_outputs.append(x)
+
+        return [semantic_outputs[i] for i in n_outputs]
+
+    @classmethod
+    def head(cls, inputs, targets, name='head', **kwargs):
+        res = []
+        for i, x in enumerate(inputs):
+            res.append(super().head(x, targets, name=name+'-'+str(i), **kwargs))
+        return res
```
SergeyTsimfer: What will happen to this list afterwards?

batchflow/models/tf/base.py
```diff
@@ -1648,74 +1648,100 @@ def output(self, inputs, predictions=None, ops=None, prefix=None, **kwargs):
...
-    def _add_output_callable(self, inputs, oper, name, attr_prefix, device, **kwargs):
+    def _add_output_callable(self, inputs, oper, name, attr_name, device, **kwargs):
         _ = kwargs
         x = oper(inputs)
         name = name or oper.__name__
-        self.store_to_attr(attr_prefix + name, x, device)
+        self.store_to_attr(attr_name, x, device)
         return x
 
+    def _add_predictions(self, predictions, name, attr_name, **kwargs):
```
SergeyTsimfer: It seems strange to always `stack` `head` outputs: what if you just wanted to have multiple outputs to process them individually in your `loss` function? That is, for example, the case for seismic noise attenuation models

batchflow/models/utils.py
```diff
@@ -5,7 +5,7 @@ def unpack_args(args, layer_no, layers_max):
     """ Return layer parameters """
     new_args = {}
     for arg in args:
-        if isinstance(args[arg], list) and layers_max > 1:
+        if isinstance(args[arg], list):
```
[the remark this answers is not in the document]
AlexeyKozhevin: OK, I will run them

batchflow/models/tf/unet.py
AlexeyKozhevin: Thanks, I've edited wrong class.

batchflow/models/tf/unet.py
AlexeyKozhevin: `kwargs` is ordinal dict

batchflow/models/tf/base.py
AlexeyKozhevin: You can use outputs separately by names `predictions_0` `predictions_1` etc.

batchflow/models/tf/unet.py
AlexeyKozhevin: I think that in that case it is better to define variable then write
```
            outputs = super().decoder(inputs[:i] + [inputs[i]] * 2, name=name,
                                      return_all=True, **{**kwargs, 'num_stages': I})
```

batchflow/models/tf/unet.py
AlexeyKozhevin: Authors proposed to implement dense blocks, I've realised it in that way.

batchflow/models/tf/unet.py
AlexeyKozhevin: It will be used to create outputs

batchflow/models/tf/base.py
AlexeyKozhevin: Previously, output-block processed multiple outputs from `head` in wrong way.
Now all outputs from `head` will be transformed by `ops` dict and in case of multiple outputs each output will have postfix `_{i}` where i is an index.

batchflow/models/tf/unet.py
roman-kh: Do not forget about `0`: `[-j-1]`

AlexeyKozhevin: I've rewritten it in last commit as 
`x = combine([x, outputs[-j-1]], **combine_args)`

PR #460, opened 2019-12-13
Few simple fixes

cdtn:
All except one are docstrings edits.

The one in code removes `Batch.load()` unnecessary `str->tuple` conversion, since it doesn't really affect `setdiff1d()` behaviour and the following `add_components()` does the same conversion.

This PR was originally aimed at fixing failing `01_batch_operations.ipynb` tutorial, but this fix is omiited now since there was a more complex one proposed. It requires more work and will be applied later.

### DIFF
--- a/batchflow/components.py
+++ b/batchflow/components.py
@@ -10,7 +10,7 @@ from .utils import is_iterable
 
 
 class AdvancedDict(dict):
-    """ dict that supports advanced indexing """
+    """ Dict that supports indexing by `list` and `np.ndarray` """
     def __getitem__(self, item):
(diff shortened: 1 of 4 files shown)


batchflow/components.py
```diff
@@ -10,7 +10,7 @@
 
 
 class AdvancedDict(dict):
-    """ dict that supports advanced indexing """
+    """ Dict that supports indexing by `list` and `np.ndarray` """
```
Dimonovez:
```suggestion
    """ Dict that supports indexing by `list` or `np.ndarray` """
```

PR #461, opened 2019-12-15
Torch improvements

SergeyTsimfer:
Changelist:

- Make `get_defaults` update dicts with `Config`-like functionality: only end leaves are updated

- `device`/`devices` are no longer saved, and `save`/`load` methods now use `dill` module for pickling

- EncoderDecoder now can use `base_model` class to create encoder

- `SideBlock` is renamed to `Branch`, and `R` now provides the same functionality as `B`

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -717,54 +717,6 @@ class TorchModel:
         return None
 
 
-    def get(self, name, from_start=False, wrap=True):
-        """ Get parts from model.
-
-        Parameters
-        ----------
-        name : str
-            Name of model part to retrieve.
-        from_start : bool
-            If True, then model is retrieved up to the part specified by `name`.
-            If False, only specified part is retrieved.
-        wrap : bool
-            Whether to wrap returned module into separate class that returns only the last output.
-        """
-        if self.model is None:
-            raise ValueError('Model is not created yet.')
-
-        order = self.full_config['order']
-
-        if name in order:
-            if from_start is False:
-                blocks = [(name, getattr(self.model, name) if hasattr(self.model, name) else None)]
-            if from_start is True:
-                index = order.index(name) + 1
-                blocks = [(item, getattr(self.model, item)) for item in order[:index] if hasattr(self.model, item)]
-
-        if name == 'encoder':
-            if hasattr(self.model.body, name):
-                encoder = self.model.body.encoder
-
-            if wrap is True and encoder.return_all is True:
-                class WrappedEncoder(nn.Module):
-                    """ Wrapper around encoder to make it return only one tensor. """
-                    def __init__(self):
-                        super().__init__()
-                        self.encoder = encoder
-
-                    def forward(self, x):
-                        return self.encoder(x)[-1]
-                encoder = WrappedEncoder()
-
-            if from_start is True:
-                index = order.index('body')
-                blocks = [(item, getattr(self.model, item)) for item in order[:index] if hasattr(self.model, item)]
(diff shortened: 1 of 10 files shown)


batchflow/models/torch/base.py
```diff
@@ -710,11 +712,59 @@ def head(cls, inputs, target_shape, classes, **kwargs):
...
+        order = self.full_config['order']
+
+        if name in order:
+            if from_start is False:
+                blocks = [(name, getattr(self.model, name) if hasattr(self.model, name) else None)]
+            if from_start is True:
+                index = order.index(name) + 1
+                blocks = [(item, getattr(self.model, item)) for item in order[:index] if hasattr(self.model, item)]
+
+        if name == 'encoder':
```
roman-kh: That's a bunch of very strange hard-code which we should get rid of

SergeyTsimfer: The original intent for this function is to exactly retrieve encoder and make it so it can be reused in other models by optionally wrapping it. Other functionality -- retrieving parts of the model  -- did come later. We can get rid of this function altogether, if we think that it is better to delegate these easy wrappings to the user

batchflow/models/torch/base.py
```diff
@@ -674,7 +676,7 @@ def initial_block(cls, inputs, **kwargs):
         torch.nn.Module or None
         """
         kwargs = cls.get_defaults('initial_block', kwargs)
-        if kwargs.get('layout'):
+        if kwargs.get('layout') or kwargs.get('base_block'):
```
nikita-klsh: Initial block congifuration  ` initial_block/filters: 128` ( example from `torch_test.ipynb`) still will not go inside this `if` block

a-arefina: What about moving `kwargs` checks inside `ConvBlock`?

batchflow/models/torch/layers/resize.py
```diff
@@ -223,7 +223,7 @@ def forward(self, x):
 
 
 
-class SideBlock(nn.Module):
+class Branch(nn.Module):
```
a-arefina: Is a module called resize a suitable place for this class?

batchflow/models/torch/base.py
```diff
@@ -1158,7 +1208,7 @@ def _add_output_op(self, inputs, oper, name, attr_prefix, **kwargs):
             output = inputs.argmax(dim=class_axis)
```
a-arefina: the line #1185 
`raise TypeError("Network output is expected to be a Tensor, but given {}".format(type(inputs)))`
should rather be 
`raise TypeError("Network output is expected to be a Tensor, but given {}".format(type(tensor)))`
because `inputs` is always a list or a tuple

batchflow/models/torch/resnet.py
```diff
@@ -78,7 +78,7 @@ def default_config(cls):
                                         pool_size=3, pool_strides=2)
 
         config['body/encoder/num_stages'] = 4
-        config['body/encoder/order'] = ['block']
+        config['body/encoder/order'] = ['skip', 'block']
```
a-arefina: Can't understand, why do you need this skip. It seems to me, that the skip is discarded in `EncoderModule.forward` because `return_all=False` is passed while creating EncoderModule in  `Encoder.body`. `return_all` isn't set in config for ResNet

SergeyTsimfer: It can be useful if one uses `ResNet18`, for example, as `base_model` for encoder-decoder.

PR #462, opened 2019-12-17
Add VNet model

### LINKED ISSUE #351: Update torch.Vnet with EncoderDecoder
Do this after #345

cdtn:
This closes #351.


batchflow/models/torch/vnet.py
```diff
@@ -0,0 +1,114 @@
...
+
+        config['body/decoder/order'] = ['upsampling', 'combine', 'block']
+        config['body/decoder/blocks'] += dict(base=ResBlock, layout=['cna'*3, 'cna'*3, 'cna'*2, 'cna'],
+                                              filters=[256, 128, 64, 32], kernel_size=5)
+        config['body/decoder/upsample'] += dict(layout='cna', filters=[128, 64, 32, 16], kernel_size=2, strides=2)
+
+        config['loss'] = 'ce'
+        return config
+
+    def build_config(self):
```
SergeyTsimfer: Current interface is really dumb: you need to set everything to `None` to get some values back. It would be much appreciated if you suggested some better way of doing it.

batchflow/models/torch/vnet.py
```diff
@@ -0,0 +1,114 @@
...
+        config['body/encoder/blocks'] += dict(base=ResBlock, layout=['cna', 'cna'*2, 'cna'*3, 'cna'*3],
+                                              filters=[16, 32, 64, 128], kernel_size=5)
+        config['body/encoder/downsample'] += dict(layout='cna', filters=[32, 64, 128, 256], kernel_size=2, strides=2)
+
+        config['body/embedding'] += dict(base=ResBlock, layout='cna'*3, filters=256, kernel_size=5)
+
+        config['body/decoder/order'] = ['upsampling', 'combine', 'block']
+        config['body/decoder/blocks'] += dict(base=ResBlock, layout=['cna'*3, 'cna'*3, 'cna'*2, 'cna'],
+                                              filters=[256, 128, 64, 32], kernel_size=5)
+        config['body/decoder/upsample'] += dict(layout='cna', filters=[128, 64, 32, 16], kernel_size=2, strides=2)
```
SergeyTsimfer:
```suggestion
        config['body/decoder/upsample'] += dict(layout='tna', filters=[128, 64, 32, 16], kernel_size=2, strides=2)
```

batchflow/models/torch/vnet.py
```diff
@@ -0,0 +1,96 @@
...
+from .blocks import ResBlock
+
+
+
+class VNet(EncoderDecoder):
+    """ VNet-like model.
+
+    Parameters
+    ----------
+    build_from_stages : bool
```
roman-kh: Do we really need this?

batchflow/tests/notebooks/torch_test.ipynb
```diff
@@ -477,6 +477,23 @@
...
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "config = {\n",
+    "    'body/encoder/num_stages': 5,\n",
+    "    'build_from_stages' : True,\n",
+    "}\n",
+    "\n",
+    "# ppl = run('segmentation', VNet, dict(), 'vnet')\n",
```
roman-kh: Remove

batchflow/models/torch/vnet.py
cdtn: Same logic is implemented in `UNet.build_config`, I just cleared it a bit, as @SergeyTsimfer asked in https://github.com/analysiscenter/batchflow/pull/462#discussion_r359054991. IMO, it's a useful param, that allows writing short configs in notebooks (i.e. `{'body/encoder/num_stages': n, 'build_from_stages': True}`), when everything we want is to increase network depth.

roman-kh: Within `build_config` one can easily check for `num_stages` within `self.config`. Which is essentially the same as having `build_from_stages ` without `build_from_stages`, isn't it?

cdtn: Not exactly. Since default and outer configs are always combined (see [`TorchModel.build`](https://github.com/analysiscenter/batchflow/blob/master/batchflow/models/torch/base.py#L313)), `num_stages` will always be in `self.config` (as well as the other default params). And therefore from inside of `build_config` it's impossible to distinguish between the case when `num_stages` is actually the only param that we mentioned in outer config (and want to build other params from it) and the case when along with `num_stages` we also pass something else (`filters`, `layout`, etc).

cdtn: Not exactly. Since default and outer configs are always combined (see [`TorchModel.build`](https://github.com/analysiscenter/batchflow/blob/master/batchflow/models/torch/base.py#L313)), the same `num_stages` value (if passed through outer config) will always present in both `self.config` and `self.full_config`. And therefore, _if we want_ to have functionality of so-called 'building from stages', we need to distinguish from inside of `build_config` between two cases:
1. `num_stages` is actually the only param that we mentioned in outer config;
2. along with `num_stages` some other params (`filters`, `layout`, etc) were mentioned.

We can achieve that by either checking for missing params in `self.config` (almost like in [earlier version](https://github.com/analysiscenter/batchflow/blob/b4c7b44ee23f4e8da5008c79327c22de1df13b2a/batchflow/models/torch/vnet.py#L77)) or via notorious `build_from_stages`.

batchflow/models/torch/vnet.py
```diff
@@ -0,0 +1,96 @@
...
+        return config
+
+    def build_config(self):
+        config = super().build_config()
+
+        if config.get('build_from_stages'):
+            num_stages = config.get('body/encoder/num_stages')
+            encoder_filters = [16 * 2**i for i in range(num_stages)]
+            encoder_layout = ['cna', 'cna'*2] + ['cna'*3] * (num_stages - 2) if num_stages != 1 else 'cna'
+            downsample_filters = [32 * 2**i for i in range(num_stages)]
```
SergeyTsimfer:
```suggestion
            downsample_filters = [16 * 2**(i + 1) for i in range(num_stages)]
```

batchflow/models/torch/vnet.py
```diff
@@ -0,0 +1,96 @@
...
+                                              filters=[256, 128, 64, 32], kernel_size=5)
+        config['body/decoder/upsample'] += dict(layout='tna', filters=[128, 64, 32, 16], kernel_size=2, strides=2)
+
+        config['loss'] = 'ce'
+        return config
+
+    def build_config(self):
+        config = super().build_config()
+
+        if config.get('build_from_stages'):
```
SergeyTsimfer: `rebuild`, maybe?
Also, now the number `16` is hardcoded; you can make it so `rebuild` allows to set the base number of filters by either passing `rebuild = 16` or `rebuild/filter_base = 16` to the config.
After finalizing your work here, do the same to `UNet`, please

batchflow/models/torch/vnet.py
```diff
@@ -0,0 +1,106 @@
...
+        config['body/decoder/upsample'] += dict(layout='tna', filters=[128, 64, 32, 16], kernel_size=2, strides=2)
+
+        config['loss'] = 'ce'
+        return config
+
+    def build_config(self):
+        config = super().build_config()
+
+        if config.get('auto_build'):
+            num_stages = config.get('auto_build/num_stages')
```
roman-kh:
```suggestion
            num_stages = config.get('auto_build/num_stages', 4)
```

PR #466, opened 2019-12-25
Research

AlexeyKozhevin:
- always use alias for `ResearchNamed Expressions`
- `load` in `__init__` of `Results`, created `pd.DataFrame` will be stored in `df` attribute
- `Results` can be filtered by experiment (`sample_index`)
- separate `experiment_path` (inner path to experiment folder) and `research_path` (path to research folder), create corresponding named expressions
- update notebooks


batchflow/named_expr.py
```diff
@@ -9,7 +9,7 @@ class _DummyBatch:
     """ A fake batch for static models """
     def __init__(self, pipeline):
         self.pipeline = pipeline
-        self.dataset = pipeline.dataset
+        self.dataset = pipeline.dataset if pipeline is None else None
```
roman-kh:
```suggestion
        self.dataset = pipeline.dataset if pipeline is not None else None
```

batchflow/research/named_expr.py
```diff
@@ -3,15 +3,15 @@
 import os
 
 from .results import Results
-from ..named_expr import NamedExpression
+from ..named_expr import NamedExpression, eval_expr
 
-class ResearchNamedExpression(NamedExpression):
+class ResearchNamedExpression(NamedExpression): #ResearchNamedExpression
```
roman-kh: `#ResearchNamedExpression` Why?

batchflow/research/results.py
```diff
@@ -14,11 +14,75 @@ class Results:
...
+        results = results[(results.iterations >= 5000) &
+                            (results.name == 'test_accuracy') & results.layout.isin(['cna', 'can'])]
+        ```
     """
-    def __init__(self, path):
+    def __init__(self, path, *args, **kwargs):
         self.path = path
         self.description = self._get_description()
         self.configs = None
+        self.df = self._load(*args, **kwargs)
```
roman-kh: Why private?

AlexeyKozhevin: Because if you want to load results again you can create a new instance, it will be the same

batchflow/research/named_expr.py
AlexeyKozhevin: For consistency :) OK, I will remove it

roman-kh: There is also an inconsistent number of spaces after `#` :-)

PR #481, opened 2020-02-10
exec init_model in pipeline before

nikita-klsh:
`init_model` are not unpacking arguments passed as `NamedExpressions`.
e.g. pipeline
```
pipeline_config = dict(model_class=Resnet18, model_name='my_model')
pipeline = (data.pipeline(pipeline_config)
                    .init_model('dynamic', C('model_class'), C('name'), model_config) 
                    .train_model(C('name'), ...)

```
Throws the error since `train_model` tries to get `my_model` model which exists under the other name in the pipeline (NamedExpression itself).

### DIFF
--- a/batchflow/once_pipeline.py
+++ b/batchflow/once_pipeline.py
@@ -142,15 +142,14 @@ class OncePipeline:
         >>> pipeline.before
               .init_model('dynamic', MyModel, config={'input_shape': C(lambda batch: batch.images.shape[1:])})
         """
-        self._add_action(INIT_MODEL_ID, _args=dict(mode=mode, model_class=model_class, model_name=name, config=config))
-        return self
+        action = dict(mode=mode, model_class=model_class, model_name=name, config=config)
+        if mode == 'static':
+            self._exec_init_model(action)
+            return self
+        return self._add_action(INIT_MODEL_ID, _args=action)
 
     def _exec_init_model(self, action):
-        name = eval_expr(action['model_name'], pipeline=self.pipeline)
-        model_class = eval_expr(action['model_class'], pipeline=self.pipeline)
-        mode = eval_expr(action['mode'], pipeline=self.pipeline)
-        config = eval_expr(action['config'], pipeline=self.pipeline)
-        self.pipeline.models.init_model(mode, model_class, name, config)
+        self.pipeline._exec_init_model(None, action)
 
     def save_model(self, name, *args, **kwargs):
         """ Save a model """
--- a/batchflow/once_pipeline.py
+++ b/batchflow/once_pipeline.py
@@ -6,7 +6,7 @@ import logging
 import numpy as np
 
 from .named_expr import NamedExpression, eval_expr
-from ._const import ACTIONS, LOAD_MODEL_ID, SAVE_MODEL_ID, UPDATE_ID
+from ._const import ACTIONS, LOAD_MODEL_ID, SAVE_MODEL_ID, UPDATE_ID, INIT_MODEL_ID
 
 
 class OncePipeline:
@@ -142,8 +142,15 @@ class OncePipeline:
         >>> pipeline.before
               .init_model('dynamic', MyModel, config={'input_shape': C(lambda batch: batch.images.shape[1:])})
         """
-        self.pipeline.models.init_model(mode, model_class, name, config=config)
+        self._add_action(INIT_MODEL_ID, _args=dict(mode=mode, model_class=model_class, model_name=name, config=config))
         return self
+
+    def _exec_init_model(self, action):
+        name = eval_expr(action['model_name'], pipeline=self.pipeline)
+        model_class = eval_expr(action['model_class'], pipeline=self.pipeline)
+        mode = eval_expr(action['mode'], pipeline=self.pipeline)
+        config = eval_expr(action['config'], pipeline=self.pipeline)
+        self.pipeline.models.init_model(mode, model_class, name, config)
 
     def save_model(self, name, *args, **kwargs):
         """ Save a model """
--- a/batchflow/tests/pipeline_init_model_test.py
+++ b/batchflow/tests/pipeline_init_model_test.py
@@ -1,55 +1,48 @@
-""" Tests for pipeline init_model action """
-# pylint: disable=import-error, no-name-in-module
-import pytest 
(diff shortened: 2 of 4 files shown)


batchflow/once_pipeline.py
```diff
@@ -142,9 +142,16 @@ def init_model(self, mode, model_class=None, name=None, config=None):
         >>> pipeline.before
               .init_model('dynamic', MyModel, config={'input_shape': C(lambda batch: batch.images.shape[1:])})
         """
-        self.pipeline.models.init_model(mode, model_class, name, config=config)
+        self._add_action(INIT_MODEL_ID, _args=dict(mode=mode, model_class=model_class, model_name=name, config=config))
         return self
 
+    def _exec_init_model(self, action):
```
roman-kh: That would break a lot of things.
Make proper tests for all modes with multiple models in the pipeline

batchflow/tests/pipeline_init_model_test.py
```diff
@@ -6,13 +8,17 @@
 
 @pytest.fixture
 def dataset():
+    """ Toy dataset """
     index = DatasetIndex(100)
     return Dataset(index, Batch)
```
roman-kh:
```suggestion
    return Dataset(100)
```

batchflow/tests/pipeline_init_model_test.py
```diff
@@ -0,0 +1,55 @@
...
+def dataset():
+    """ Toy dataset """
+    index = DatasetIndex(100)
+    return Dataset(index, Batch)
+
+@pytest.mark.parametrize('use_algebra', [True, False])
+@pytest.mark.parametrize('config', [dict(mode='static', base_class=BaseModel, name='model', config={}),
+                                    dict(mode='dynamic', base_class=BaseModel, name='model', config={})])
+def test_single_model(use_algebra, config, dataset):
+    """ Verifies that the model in init_model is initialized as expected with config provided, i.e.
```
roman-kh: This test does not test anything. The model is never addressed and never used in training or inference.

batchflow/tests/pipeline_init_model_test.py
```diff
@@ -0,0 +1,55 @@
...
+    if not use_algebra:
+        pipeline = dataset.pipeline(config).init_model(C('mode'), C('base_class'), C('name'), C('config'))
+    else:
+        pipeline = dataset.pipeline().init_model(C('mode'), C('class'), C('base_name'), C('config')) << config
+    pipeline.next_batch(1)
+
+    if config['mode'] == 'static': 
+        assert type(pipeline.m(config['name']) == config['class']
+    else:
+        assert type(pipeline.m(config['name']) == NonInitializedModel
```
roman-kh: `is` or `isinstance` fits the purpose
if a model is NonInitializedModel then it was not initialized apparently

batchflow/tests/pipeline_init_model_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+@pytest.mark.parametrize('use_algebra1', [True, False])
```
roman-kh: use_algebra1 and 2 are extremely non obvious.Give clearer names

batchflow/tests/pipeline_init_model_test.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.mark.parametrize('use_algebra2', [True, False])
+@pytest.mark.parametrize('mode', ['static', 'dynamic'])
+@pytest.mark.parametrize('config1', 'config2' [(dict(class=BaseModel, name='model1', config={}),
+                                                dict(class=BaseModel, name='model2', config={})])
+def test_multiple_models(dataset, use_algebra1, use_algebra2, mode,  config1, config2):
+    """ Verifies that several modeles can be initialized and stored in the pipeline via sequentially linking
+    different configs to the pipeline. """
+    config1.update(mode=mode)
```
roman-kh: all what is going on below is extremely confusing

batchflow/tests/pipeline_init_model_test.py
nikita-klsh: I made a bit messy. We are not interested here in addressing the model but storing

batchflow/tests/pipeline_init_model_test.py
nikita-klsh: > all what is going on below is extremely confusing

The tests provided essentially checks two things, which are currently not working in `master` 

1.  Args passed to `init_model` as C() expressions are being properly unpacked inside the method.
2. Sequentially linking configs with different models to the same pipeline with `init_model` method results in properly storing all the models.

batchflow/tests/pipeline_init_model_test.py
nikita-klsh: > This test does not test anything. The model is never addressed and never used in training or inference.

I made a bit messy. We are not interested here in addressing the model but storing

batchflow/tests/pipeline_init_model_test.py
roman-kh: If the code needs explanation it has to be rewritten. No explanation is needed besides the code.

batchflow/tests/pipeline_init_model_test.py
roman-kh: There is no point in storing anything if you never address it in a proper way. You can store it in /dev/null then.

Moreover, you killed the whole idea of static models without even realizing it.

Furthermore, `pipeline.init_model` and `pipeline.before.init_model` now works differently.

Besides, you evaluate named expressions in a peculiar fashion which is also very different from the rest of `Pipeline`.

batchflow/tests/pipeline_init_model_test.py
```diff
@@ -0,0 +1,48 @@
...
+    if not use_pipeline_algebra:
+        pipeline = dataset.pipeline().init_model('static', BaseModel, C('name'), {}) << config1
+    else:
+        pipeline = dataset.pipeline(config1).init_model('static', BaseModel, C('name'), {})
+    pipeline.next_batch(1)
+
+    pipeline = pipeline << config2
+    pipeline.next_batch(1)
+
+    assert pipeline.m(config1['name']) and pipeline.m(config2['name'])
```
roman-kh: name and name are confusing. which name is here?

PR #491, opened 2020-03-05
Fix sampler

SergeyTsimfer:
This PR proposes to

- store constant values in instance of `HistoSampler` instead of computing them at every `sample` call

- Adds time of `eval_expr` of arguments for every action in `Pipeline` to the profiling info


batchflow/pipeline.py
```diff
@@ -764,16 +764,16 @@ def _exec_all_actions(self, batch, actions=None):
         actions = actions or self._actions
 
         for action in actions:
+            if self._profile:
+                start_time = time.time()
+                self._profiler.enable()
```
roman-kh: Perhaps it might be convenient to split arg evaluation and action execution for profiling puproses

batchflow/sampler.py
```diff
@@ -495,7 +496,15 @@ def __init__(self, histo=None, edges=None, seed=None, **kwargs):
         else:
             raise ValueError('Either `histo` or `edges` should be specified.')
 
+        self.l_all = cart_prod(*(range_dim[:-1] for range_dim in self.edges))
+        self.h_all = cart_prod(*(range_dim[1:] for range_dim in self.edges))
+
+        self.probs = (self.bins / np.sum(self.bins)).reshape(-1)
+        self.nonzero_probs_idx = np.asarray(self.probs != 0.0).nonzero()[0]
```
akoryagin: Do you really need both `self.probs != 0.0` and `.nonzero()`?

PR #492, opened 2020-03-12
Make `@apply_transform` great

cdtn:
Move all `MethodsTransformingMeta` logic to `@apply_transform` with a help of little argument trick.

### DIFF
--- a/batchflow/batch_image.py
+++ b/batchflow/batch_image.py
@@ -605,8 +605,8 @@ class ImagesBatch(BaseImagesBatch):
                 background = np.zeros(original_shape, dtype=np.uint8)
             else:
                 background = np.zeros((*original_shape, n_channels), dtype=np.uint8)
-            return self._put_on_background_(transformed_image, background, origin)
-        return self._crop_(transformed_image, origin, original_shape, True)
+            return self.put_on_background(transformed_image, background, origin)
+        return self.crop(transformed_image, origin, original_shape, True)
 
     @apply_transform
     def filter(self, image, mode, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


SergeyTsimfer:
you-did-it-24a783eb67

batchflow/batch_image.py
```diff
@@ -605,8 +603,8 @@ def _preserve_shape(self, original_shape, transformed_image, origin='center'):
                 background = np.zeros(original_shape, dtype=np.uint8)
             else:
                 background = np.zeros((*original_shape, n_channels), dtype=np.uint8)
-            return self._put_on_background_(transformed_image, background, origin)
-        return self._crop_(transformed_image, origin, original_shape, True)
+            return self.put_on_background(transformed_image, background, origin, apply_transform=False) # pylint: disable=unexpected-keyword-arg
```
roman-kh: Using pylint disable make the code verbose, clumsy and inconvenient

cdtn:
See no way to design it better. For now, the choices are:

1. Use metaclass to make two versions of the same method (one wrapped with `apply_transform` and one doesn't) and stick to corresponding batch methods calling convention (that's the current implementation in [master](https://github.com/analysiscenter/batchflow/tree/6308531fa8ebec06f477762189a30b0ebfbba414)). It's reliable and mostly explicit, but not the easiest one to understand at first glance.
2. Parse `traceback.format_stack()`, trying to understand where the method call came from (see the [first commit to this PR](https://github.com/analysiscenter/batchflow/pull/492/commits/f237593b71bd15c365a01b57605a6b803d00ca74)). The approach is unreliable since not fully predictable.
3. Pass an extra argument when calling one batch method from inside of the other one (and that is the [second commit to this PR](https://github.com/analysiscenter/batchflow/pull/492/commits/f91f7c94b86766d03253ed4b184df45ae8e5b4c8)). Since the argument is missing in the original method header (and being intercepted with decorator) pylint raises an error and shutting it up looks verbose, clumsy and inconvenient.

Closed until better times.

PR #494, opened 2020-03-25
Fix zero gpu allocation

AlexeyKozhevin:
Fix allocation of GPU:0 when visible devices are not specified.
NOTE: For `tensorflow>=1.15.0`, otherwise GPU:0 also will be allocated when model has `conv2d`

### DIFF
--- a/batchflow/research/workers.py
+++ b/batchflow/research/workers.py
@@ -175,12 +175,12 @@ class PipelineWorker(Worker):
         i, job = self.job
         n_branches = len(job.configs)
         mapping = {
-            item: i
-            for i, item in enumerate({self.devices[i]['device'] for i in range(n_branches)})
+            item: 'gpu:' + str(i)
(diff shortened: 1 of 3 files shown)


batchflow/research/workers.py
```diff
@@ -174,7 +174,13 @@ def init(self):
         """ Run before job execution. """
         i, job = self.job
         n_branches = len(job.configs)
-        self.device_configs = [self.devices[i] for i in range(n_branches)]
+        mapping = {
+            item: i
+            for i, item in enumerate({self.devices[i]['device'] for i in range(n_branches)})
+        }
+        os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"
+        os.environ["CUDA_VISIBLE_DEVICES"] = ','.join([self.devices[i]['device'] for i in range(n_branches)])
```
a-arefina: in n_branches > number of real devices, there will be repeating values, right? Is it OK?

AlexeyKozhevin: Yes

PR #495, opened 2020-03-27
Test loss

a-arefina:
Make test loss great again!

This PR fixes  loss and metrics calculation in research pipelines added with run=True


a-arefina:
Some notebooks in tutorials need to be recalculated, but wait for #494

batchflow/models/metrics/loss.py
```diff
@@ -0,0 +1,44 @@
+"""  Loss as a Metrics to be used in research pipelines added with `run=True` """
+
+import numpy as np
+
+from .base import Metrics
+
+
+class Loss(Metrics):
```
AlexeyKozhevin: Can we use `RegressionMetrics` instead of the new `Loss` class?

a-arefina: We can't use any existing metric because they all calculate needed values from given predictions and targets, and loss is already calculated during prediction. 

We also have to keep batch lengths along with average loss on each batch to compute true average loss on the whole dataset. These are the motivations for a new class.

examples/tutorials/research/02_advanced_usage_of_research.ipynb
```diff
@@ -57,7 +57,7 @@
    "source": [
```
AlexeyKozhevin, line 57: <p><code>os.environ["CUDA_VISIBLE_DEVICES"] = "1, 2"</code>  is unnecessary now</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/analysiscenter/batchflow/pull/495/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

batchflow/models/metrics/loss.py
```diff
@@ -0,0 +1,44 @@
...
+        self._agg_fn_dict.update(mean=agg_loss)
+
+        def batchwise_loss(args):
+            losses, _ = args
+            return losses
+
+        self._agg_fn_dict.update(batchwise=batchwise_loss)
+
+    def append(self, metrics):
+        """ Extend with data from another metrics"""
```
AlexeyKozhevin, line 39: Add `.` and whitespace

batchflow/tests/test_loss_test.py
```diff
@@ -0,0 +1,49 @@
...
+from batchflow import Dataset, B, V
+from batchflow.research import Research
+
+
+np.random.seed(2020)
+NUM_ITEMS = 10
+DATA = np.random.sample(NUM_ITEMS)
+
+
+@pytest.mark.skip(reason="check pytest falure in github actions")
```
AlexeyKozhevin, line 16:
```suggestion
@pytest.mark.skip(reason="check pytest failure in github actions")
```

a-arefina:
[make test loss great again!](https://trello.com/c/8dLvC90E/36-make-test-loss-great-again)

PR #497, opened 2020-04-07
refractor opensets + coco segmentation dataset

nikita-klsh:
Current datasets have the following issues:

1. Datasets itself are not being indexed (you can't iterate over it in the pipelines)
2. `dataset.images`/`dataset.labels` contains `dataset.train`/`dataset.test` parts instead of desired content.
3. Calling `split` several times does nothing as you always tied to the public train/test split.

This PR suggests slight modifications for `Openset` class and all available datasets to overcome these issues. Besides, it extends functionality to work not only with datasets loaded directly into RAM but those which files read from the disk.

It also contains a COCO dataset for semantic segmentation.


batchflow/opensets/base.py
```diff
@@ -9,11 +8,14 @@ class Openset(Dataset):
...
-            preloaded = self.download(path=path)
-        super().__init__(index, batch_class, preloaded=preloaded, **kwargs)
-        if train_test:
-            self.split()
+
+        if index is None:
+            preloaded, index = self.download(path=path)
+        super().__init__(index, batch_class=batch_class, preloaded=preloaded, **kwargs)
+
+        if self._train_index and self._test_index:
```
SergeyTsimfer, line 16: The fact that you explicitly return `preloaded` and `index` from `download` method, and implicitly set `self._train_index` and `self._test_index` during the same call still makes me uncomfortable

batchflow/opensets/cifar.py
```diff
@@ -71,9 +71,15 @@ def _gather_extracted(all_res):
             test_data = _gather_extracted(all_res)
         logger.info("Extracted")
 
+        images = np.concatenate([train_data[0], test_data[0]])
+        labels = np.concatenate([train_data[1], test_data[1]])
+
         self._train_index = DatasetIndex(np.arange(len(train_data[0])))
-        self._test_index = DatasetIndex(np.arange(len(test_data[0])))
-        return train_data, test_data
+        self._test_index = DatasetIndex(np.arange(len(train_data[0]), len(train_data[0]) + len(test_data[0])))
```
SergeyTsimfer, line 78: Giving meaningful names like `test_len`, `train_len`, `total_len` would both increase readability and is reusable

batchflow/opensets/coco.py
```diff
@@ -0,0 +1,112 @@
+""" Datasets for COCO challenge tasks, http://cocodataset.org/#home.
+    Yet contains only the dataset for Semantic Segmentation. """
```
SergeyTsimfer, line 2:
```suggestion
    Currently contains only the dataset for Semantic Segmentation. """
```

batchflow/opensets/coco.py
```diff
@@ -0,0 +1,112 @@
...
+    TRAIN_IMAGES_URL = 'http://images.cocodataset.org/zips/train2017.zip'
+    TEST_IMAGES_URL = 'http://images.cocodataset.org/zips/val2017.zip'
+
+    def __init__(self, *args, preloaded=None, **kwargs):
+        super().__init__(*args, preloaded=preloaded, **kwargs)
+
+
+    @parallel(init='_get_from_urls', post='_post_fn', target='t')
+    def download(self, url, folder, train_val, path=None):
+        """ Download the archives and extracts it's content. Downloading is performed in parallel manner.
```
SergeyTsimfer, line 33:
```suggestion
        """ Download the archives and extract their contents in a parallel manner.
```

batchflow/opensets/coco.py
```diff
@@ -0,0 +1,112 @@
...
+        filename = basename(url)
+        localname = os.path.join(path, filename)
+        if not os.path.isfile(localname):
+            r = requests.get(url, stream=True)
+            file_size = int(r.headers['Content-Length'])
+            chunk = 1
+            chunk_size = 1024 * 1000 #MBs
+            num_bars = int(file_size / chunk_size)
+            with open(localname, 'wb') as f:
+                for chunk in tqdm.tqdm(r.iter_content(chunk_size=chunk_size), total=num_bars,
```
SergeyTsimfer, line 49: Add option for using/not using the progress bar

batchflow/opensets/coco.py
```diff
@@ -0,0 +1,112 @@
...
+                       ['COCOImages', 'COCOImages', 'COCOMasks', 'COCOMasks'],
+                       ['train2017', 'val2017', 'train2017', 'val2017'])
+        return [[url, folder, train_val] for url, folder, train_val in iterator]
+
+    def _extract_archive(self, localname, extract_to):
+        with ZipFile(localname, 'r') as archive:
+            archive.extractall(extract_to)
+
+    def _extract_if_not_exist(self, localname, folder, train_val):
+        """ Extracts the arcive to the specific folder. Returns the path to this filder"""
```
SergeyTsimfer, line 85: typo

batchflow/opensets/coco.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @property
+    def _get_from_urls(self):
+        """ List of URL to download from, folder where to extract, and indicator whether its train or val part. """
+        iterator = zip([self.TRAIN_IMAGES_URL, self.TEST_IMAGES_URL, self.MASKS_URL, self.MASKS_URL],
+                       ['COCOImages', 'COCOImages', 'COCOMasks', 'COCOMasks'],
+                       ['train2017', 'val2017', 'train2017', 'val2017'])
+        return [[url, folder, train_val] for url, folder, train_val in iterator]
+
+    def _extract_archive(self, localname, extract_to):
```
SergeyTsimfer, line 80: Do you really need this function? It is used only once

batchflow/opensets/coco.py
```diff
@@ -0,0 +1,112 @@
...
+        with ZipFile(localname, 'r') as archive:
+            archive.extractall(extract_to)
+
+    def _extract_if_not_exist(self, localname, folder, train_val):
```
SergeyTsimfer, line 84: Do you really need this function? Looks like it would be exactly the same for any other `CoCo` datasets (if we were to add more), and it would be way more clear if the contents of that function were part of the `download` method of `BaseCOCO` class
And you won't need that strange return of path from an extracting function

batchflow/opensets/coco.py
```diff
@@ -0,0 +1,112 @@
...
+            chunk_size = 1024 * 1000 #MBs
+            num_bars = int(file_size / chunk_size)
+            with open(localname, 'wb') as f:
+                for chunk in tqdm.tqdm(r.iter_content(chunk_size=chunk_size), total=num_bars,
+                                       unit='MB', desc=filename, leave=True):
+                    f.write(chunk)
+        return self._extract_if_not_exist(localname, folder, train_val)
+
+
+class COCOSegmentation(BaseCOCO):
```
SergeyTsimfer, line 55: As far as I recall, it was you who added names for each of the classes in `Pascal` dataset, so why would not you do the same for `CoCo`?

batchflow/opensets/base.py
nikita-klsh: That's how things were done before, with only difference having `train_test` flag. 
All the stuff regarding `preloaded`, `index`, `train_index` and `test_index` must be inferred inside `download` method, so potentially it could return 4 values. I found this a bit overwhelming. 
Would like to hear your opinion.

SergeyTsimfer: This function is ~nevert used be the end user of the `Openset`, so I don't see any problems with it returning as many values as is convenient
A somewhat related example: current `tensorflow.Dataset` classes return tuple of `((train_x, train_y), (test_x, test_y))`

roman-kh:
Is it ready to merge?

batchflow/opensets/base.py
```diff
@@ -27,13 +27,12 @@ def download(self, path):
...
-    def split(self, shares=0.8, shuffle=False):
-        if self.train_test:
-            train_data, test_data = self.preloaded  # pylint:disable=unpacking-non-sequence
-            self.train = type(self)(self._train_index, self.batch_class, train_test=False, preloaded=train_data)
-            self.test = type(self)(self._test_index, self.batch_class, train_test=False, preloaded=test_data)
-        else:
-            super().split(shares, shuffle)
+    def _infer_train_test_index(self, train_len, test_len):
+        total_len = train_len + test_len
+        index = DatasetIndex(list(range(total_len)))
```
SergeyTsimfer, line 32: As far as I know, `DatasetIndex` can be initialized with an integer with the same result

PR #504, opened 2020-04-27
Trackers

SergeyTsimfer:
This PR proposes to add:

- monitoring tools (namely, new `ResourceMonitor` class, as well as context managers for conveniency), for example
```
with monitor_resource('gpu', frequency=0.5) as monitor:
    # train model
```
to better understand resource (cpu, gpu, memory) utilization

- tracking tools (namely, new `Notifier` class) to provide better utilization of `tqdm` progress-bars in `Pipeline`, as well as to plot graphs of, for example, loss values, dynamically during model training

- `Notifier` class is also capable of using any of the resource monitoring utilities provided by `ResourceMonitor`s. Old functionality of just passing `n`/`True` is working too

image


batchflow/notifier.py
```diff
@@ -0,0 +1,222 @@
...
+from tqdm.auto import tqdm as tqdm_auto
+
+import numpy as np
+from IPython import display
+import matplotlib.pyplot as plt
+
+from .monitor import ResourceMonitor, MONITOR_ALIASES
+from .named_expr import NamedExpression, eval_expr
+
+class Notifier:
```
roman-kh, line 16: Let's make a bit more structured class hierarchy like:
- Notifier
---- Bar
---- Monitor
------ GPU_Monitor
etc

and Notifer can also take a list of other notifiers and update them at once

batchflow/pipeline.py
```diff
@@ -1239,13 +1240,13 @@ def rebatch(self, batch_size, fn=None, components=None, batch_class=None):
...
 
-    def _put_batches_into_queue(self, gen_batch, bar, bar_desc):
+    def _put_batches_into_queue(self, gen_batch, bar):
         while not self._stop_flag:
             self._prefetch_count.put(1, block=True)
             try:
                 batch = next(gen_batch)
                 if bar:
-                    update_bar(bar, bar_desc, pipeline=self, batch=batch)
+                    bar.update(pipeline=self, batch=batch)
```
roman-kh, line 1249: there is no bar anymore, only notifiers

batchflow/pipeline.py
```diff
@@ -1239,13 +1240,13 @@ def rebatch(self, batch_size, fn=None, components=None, batch_class=None):
         return new_p._add_action(REBATCH_ID, _args=dict(batch_size=batch_size, pipeline=self, fn=fn,
                                                         components=components, batch_class=batch_class))
 
-    def _put_batches_into_queue(self, gen_batch, bar, bar_desc):
+    def _put_batches_into_queue(self, gen_batch, bar):
         while not self._stop_flag:
             self._prefetch_count.put(1, block=True)
             try:
                 batch = next(gen_batch)
                 if bar:
```
roman-kh, line 1248: Let's always create notifier and always call `notifier.update()`... sometimes it will do nothing

batchflow/monitor.py
```diff
@@ -0,0 +1,190 @@
...
+
+class ResourceMonitor:
+    """ Periodically runs supplied function in a separate process and stores its outputs.
+
+    Parameters
+    ----------
+    function : callable
+        Function to use. If not provided, defaults to the `get_usage` static method.
+    frequency : number
+        Periodicity of function calls.
```
AlexeyKozhevin, line 26:
```suggestion
        Periodicity of function calls (seconds).
```

batchflow/dsindex.py
```diff
@@ -7,7 +7,7 @@
 import numpy as np
 
 from .base import Baseset
-from .utils import create_bar, update_bar
```
AlexeyKozhevin, line 10: Also remove them from `utils.py`

batchflow/notifier.py
SergeyTsimfer: What is the reason for it? Currently, `ResourceMonitor` has nothing to do with notifications and the class is used to generate some data, while `Notifier` is used to display it, so that class hierarchy would be rather strange.

As for the `Notifier` being able to initialize from a list of previously created `Notifier`s: should it be a creation of a new instance that tracks the same monitors/pipeline variables or just storing all the passed notifiers and clicking `update` on them at the same time when `self` is `update`d? While the second option is very easy to implement, what is the use case for it? The first option is way harder and would require making some decisions; it would also make the code hard to read: the main problem is what to do with `bar`s from each from the passed `Notifier`. Considering the fact that bar creation is already the hardest part of the proposed code, in either way it would render a (currently) very simple and clear class completely unreadable.

With all that, my question is: what is the benefit/use case of going all this trouble?

batchflow/notifier.py
```diff
@@ -0,0 +1,267 @@
...
+        if add_suptitle:
+            title = self.format_meter(self.n, self.total, time()-self.start_t, ncols=80)
+            plt.suptitle(title, y=0.99, fontsize=14)
+        plt.show()
+
+    # Convenient alias for working inspect instance after pipeline run
+    visualize = update_plots
+
+
+    def __call__(self, iterable):
```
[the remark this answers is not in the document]
SergeyTsimfer: image

SergeyTsimfer:
As per @a-arefina request, I've added capability of logging data to a supplied txt file: you can do that on the fly by providing `file` initialization argument, or afterwards by using `to_file` method of `Notifier`. This feature is indeed very nice to have as it allows to seamless logging of train pipelines and variables.

I have also added the capability to save the visualization of a `Notifier` to disk: that is definitely better than cutting parts of a screenshot:D

@roman-kh  do we need a separate demo notebook for notifiers?

batchflow/notifier.py
```diff
@@ -248,11 +266,37 @@ def update_plots(self, add_suptitle=False):
...
     # Convenient alias for working inspect instance after pipeline run
     visualize = update_plots
 
+    def update_file(self):
+        with open(self.file, 'a+') as f:
+            timestamp = strftime("%Y-%m-%d  %H:%M:%S", self.timestamps[-1])
+            msg = f'Iteration {self.bar.n:5};    {self.bar.desc[:-1]}'
+            print(f'{timestamp}     {msg}', file=f)
+
+    def to_file(self, file):
```
a-arefina, line 283: This method doesn't reuse any code that prints logs on the fly. This can lead to differences between reports that are generated online and those printed upon request.

It might be more convenient to use `logging` standard library. It allows to define consistent log format, including timestamps. One can also choose different handlers to control the destination of log output. 

writing to a `logger.FileHandler` in `update_file` would be an equivalent of current `to_file`, and using a stream and flushing it would produce online logging.

IMO, there is no need to have a logger that prints information strictly on each update and this can be also very slow. Having a FileHandler that controls output to a file would be sufficient.

a-arefina:
Please also note that some tests are failing in this PR.
You can find rebatch_test fix in PR #515

batchflow/notifier.py
SergeyTsimfer: I am well aware of the `logging` module, yet it requires some care to use.
The reason why I chose not to use it here is because our `FileLogger` in `Research` collects all the info from all the loggers and puts it into `research.log` file. Furious spam from the iteration-wise notifiers from multiple processes would make this file completely unreadable; that is why I decided to use simple file writes, provided we don't need any extra functionality (for now).

SergeyTsimfer: As for the not reusing code: it hurts me to say it, but `premature generalization is the root of all evil`, as one man (almost) said. The effort required to generalize similar 3 lines of code in different places may just make it harder to read and lead to more confusion, not clarity.

a-arefina: > I am well aware of the `logging` module, yet it requires some care to use.
> The reason why I chose not to use it here is because our `FileLogger` in `Research` collects all the info from all the loggers and puts it into `research.log` file. Furious spam from the iteration-wise notifiers from multiple processes would make this file completely unreadable; that is why I decided to use simple file writes, provided we don't need any extra functionality (for now).

As for mixing logs with `Research` - you can just use another logger. 
As for readability: most of log files are unreadable. What you do in these logs is printing some info related to performance - this kind of data is extremely hard to understand simply by looking at a text file. You will always parse log files to build plots or to make some aggregation, so what you really need is a consistent logging format. You simply dump all data you have in a log in such a manner, that it is easy to extract what you really want to get automatically.

a-arefina: > As for the not reusing code: it hurts me to say it, but `premature generalization is the root of all evil`, as one man (almost) said. The effort required to generalize similar 3 lines of code in different places may just make it harder to read and lead to more confusion, not clarity.

Well, Donald Knuth actually said nothing about evils of generalization. But the evils of duplicated code are well-known. 

The next time you need to make a change in the logging process you will have to implement it twice, and you are always at risk that you make it differently in 2 cases. Even now, are you absolutely sure that 2 different logging modes produce similar results? 

Of course you can delay the refactoring until that next time, but will not it be better to design it an a way that allows easy and reliable modifications now, that you see all context?

This particular case with logs is trivial, and it is really up to you whether to refactor it or not. But for a more important feature I would strongly insist on producing code that can be modified without pain from the very beginning.

SergeyTsimfer: You are right, so I have done the refactoring requested.
As far as the logging, I still think that this class does not need the fully-fledged logger, where simple `print` to file suffices.
Also, I do love f-strings. Is there anything in the way of us updating to `Python 3.6+`, @roman-kh?

a-arefina: > Also, I do love f-strings. Is there anything in the way of us updating to `Python 3.6+`, @roman-kh?
+1 for f-strings!

batchflow/notifier.py
```diff
@@ -0,0 +1,315 @@
...
+            data_generators.extend(monitors)
+            self.has_monitors = True
+        else:
+            self.has_monitors = False
+
+        # Prepare file log
+        self.file = file
+        if self.file:
+            with open(self.file, 'w') as _:
+                pass
```
a-arefina, line 109: You're just opening and closing the file, right? What for? Am I missing something?

SergeyTsimfer: To clear the file contents (note the `w` mode)

batchflow/notifier.py
```diff
@@ -0,0 +1,315 @@
+""" Progress notyfier. """
```
AlexeyKozhevin, line 1: Notifier

batchflow/notifier.py
roman-kh: Cheers for f-string.

roman-kh (approved):

PR #506, opened 2020-04-29
Validator

AlexeyKozhevin:
Wrapper for models to unify training and validation process.


batchflow/models/validator/README.md
```diff
@@ -0,0 +1,59 @@
+# Validator
+
+Validator is an instrument to unify the process of model training and validation. To make model API, you need:
```
roman-kh, line 3: Better structure is needed here.

Start with describing the reason, then provide a very basic usage example, then more advanced examples, then how to write all the classes and methods.

batchflow/models/validator/README.md
```diff
@@ -0,0 +1,59 @@
...
+- Define `inference` method with required arguments `dataset_path` and optional argument `model_path`. `model_path` will be the path to pretrained model (if `pretrained` is defined in  `validator.yaml`) or the path which is the same as `model_path` in `train` method. In `inference` method you must define `self.predictions` and `self.targets` that will be numpy arrays.
+- Define your own metrics with
+
+## `validator.yaml`
+Validator is used to determine the whole procedure of validation.
+Let’s consider an example
+```yaml
+- Pascal Segmentation: # the task name (you can have different tasks with different models in your repo
+       - PascalModel: # name of the validator
+            class: PascalModel # validator from model_api.py
```
roman-kh, line 19: Sj we have validator and validator from model_api... confusing

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self.pretrained = config.get('pretrained')
+        self.validate_dataset = self.config['validate']['dataset']
+        self.metrics = self.config['validate']['metrics']
+        self.custom_metrics = self.config['validate'].get('custom_metrics')
+
+        self.custom_metric_values = {}
+        self.metric_values = {}
+
+    @abstractmethod
```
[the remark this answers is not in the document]
AlexeyKozhevin: Our aim is to clarify model lifecycle. I think that `train` is required even when we use pre-trained model.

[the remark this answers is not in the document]
roman-kh: And what does train do when there is a pretrained model?

batchflow/models/validator/validator.py
```diff
@@ -59,20 +59,16 @@ def inference(self, *args, **kwargs):
...
-        print("Task: {}. ModelAPI: {}".format(self.task, self.name))
-        print("Call init...")
         self.init()
 
         if self.pretrained is None:
-            print('Start train...')
             self.train(self.train_dataset, self.model_path)
 
         if self.pretrained:
             model_path = self.pretrained['model']
```
roman-kh, line 68: perhaps `__init__` is a more suitable place for this `if`

batchflow/models/validator/validator.py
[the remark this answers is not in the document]
AlexeyKozhevin: We can distinguish two aims: on the one hand, API for model (in that case we can allow to skip `train` method), on the other hand, we can check that model satisfy our requirements. In the second case we provide additional checking script were we will automatize all checks that we can do.

AlexeyKozhevin:
Refactoring: add data and model loaders as separate functions, add `kwargs` loaded from config, remove abstarct methods. Update README.md

batchflow/models/validator/README.md
```diff
@@ -0,0 +1,224 @@
...
+      class: PascalValidator
+      train:
+      pretrained:
+        path: /path/to/model
+        device: cuda:0
+      test:
+        metrics:
+        - classification:
+            evaluate:
+              - accuracy
```
roman-kh, line 130: This is extremely overcomplicated. Make it `metrics: accuracy` on the first nested level (not the sixth). You can lookup for the metrics in the metrics class and if there is a name collision throw an exception. 

`metrics: classification.accuracy` is also possible (also, not instead of)

batchflow/models/validator/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+The same as `train_loader` but for `test`.
+
+### `load_model(self, path=None, **kwargs)`
+
+Loader for pretrained model. Let's make example above more complex:
+
+```yaml
+- Pascal Segmentation:
+  - PascalValidator:
```
roman-kh, line 120: The second level os nesting may be omitted as well as the first. Let's make things simpler. Each yaml file for one model. One can have as many yaml files as he needs.

batchflow/models/validator/README.md
```diff
@@ -0,0 +1,224 @@
+# Validator
```
roman-kh, line 1: Add this documentation to \docs as rst

batchflow/models/validator/README.md
roman-kh: `---` can also be used to separate different validation definitions within one yaml file

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,284 @@
+""" API for models and validator """
```
roman-kh, line 1: Add some tests

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def run(self):
+        """ Run validator """
+        if 'train' in self.config:
+            self.train_dataset = self.train_loader(**self._train_ds)
+            self.from_train = self.train(self.train_dataset, **self.config['train'])
+        elif 'pretrained' in self.config:
+            self.from_train = self.load_model(**self._pretrained)
+
+        if 'test' in self.config:
```
roman-kh, line 159: if not, it should throw a warning

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,284 @@
...
+        for task in self.config:
+            task_name, task_config = key_value(task)
+            self.results[task_name] = {}
+            for model in task_config:
+                model_name, config = key_value(model)
+                validator = eval('model_api.'+config['class'])(config, model_name, task_name) #pylint: disable=eval-used
+                self.results[task_name][model_name] = validator
+                validator.run()
+
+    def _metrics(self, custom_metrics=False):
```
roman-kh, line 232:
```suggestion
    def _metrics(self, custom=False):
```

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def start(self):
+        """ Start validator """
+        import model_api #pylint: disable=unused-import,import-error
+        for task in self.config:
+            task_name, task_config = key_value(task)
+            self.results[task_name] = {}
+            for model in task_config:
+                model_name, config = key_value(model)
+                validator = eval('model_api.'+config['class'])(config, model_name, task_name) #pylint: disable=eval-used
```
roman-kh, line 228: There will be no need for eval with one class (without ModelAPI)

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,284 @@
...
+                            <classification_kwarg_1>: <value_1>
+                            ...
+                            evaluate:
+                                - <metric_name_0>:
+                                    <metric_kwarg_0>: <value_0>
+                                    <metric_kwarg_1>: <value_1>
+                                - <metric_name_1>:
+                                    ...
+                                ...
+                    custom_metrics: (optional, metrics defined in ModelAPI child-class)
```
roman-kh, line 200: Do not split metrics and custom metrics

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,284 @@
...
+            task_name, task_config = key_value(task)
+            self.results[task_name] = {}
+            for model in task_config:
+                model_name, config = key_value(model)
+                validator = eval('model_api.'+config['class'])(config, model_name, task_name) #pylint: disable=eval-used
+                self.results[task_name][model_name] = validator
+                validator.run()
+
+    def _metrics(self, custom_metrics=False):
+        attr = 'custom_metric_values' if custom_metrics else 'metric_values'
```
roman-kh, line 233: Why is this hardcode needed?

batchflow/models/validator/validator.py
AlexeyKozhevin: As I said before, classes can be without `train` or `inference` but we will use another one script to check whether class satisfy requirements or not.

batchflow/models/validator/README.md
AlexeyKozhevin: I think that in my `yaml` structure it's easier to understand sense of each parameter because we have two groups of parameters: for `init` and for `evaluate`. 

But I agree that something like `metrics: accuracy` is clearer. I will try to do something like that.

batchflow/models/validator/validator.py
roman-kh: Why is validator needed when there is no validation?

batchflow/models/validator/README.md
AlexeyKozhevin: I like idea with multiple files.

batchflow/models/validator/validator.py
AlexeyKozhevin: We can use `ModelAPI` just as wrapper for model. For example, it can be used to train and dump model.

batchflow/models/validator/README.md
AlexeyKozhevin: @roman-kh what do you think about that?

Example:
```
train:
    dataset: /data/train
test:
    dataset: /data/test
    metrics:
        - f1_score
        - accuracy
f1_score:
    class: classification
    axis: 1
    fmt: 'proba'
    num_classes: 18
    evaluate:
        agg: mean
        multiclass:
```

In `test` we list metrics to compute. If `ModelAPI` class has a method with the same name, it will be computed as a custom metric (`accuracy` here). Otherwise, it will be computed as BatchFlow metric and parameters of that metric we define in additional section (`f1_score`).

batchflow/models/validator/validator.py
roman-kh: Why do we need a wrapper for a model when we have a model class?

Let's make it dead simple:
- Validator class is to validate a model
- one should define a child class with specific load, save, train, predict and custom metrics methods

LIghtningValidator can also be an example of a child Validator class.

batchflow/models/validator/README.md
roman-kh: This look great. However, `f1_score` requires too many parameters. Good defaults are needed here in order to eliminate these parameters for most cases

roman-kh: Remove additional nesing for evaluate. 1 level is more than enough for the metrics

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,172 @@
+""" API for models and validator """
+
+import warnings
+
+import yaml
+
+from ..metrics import ClassificationMetrics, SegmentationMetricsByPixels, \
+                      SegmentationMetricsByInstances, RegressionMetrics
+
+METRICS = dict(
```
roman-kh, line 10: there is `batchflow.pipeline.METRICS`. You can use it (and even move to batchflow namepsace)

batchflow/models/validator/README.md
AlexeyKozhevin: As for defaults, I think that defaults should be from BatchFlow metric classes. If you can use defaults from BatchFlow, you don't need to configure that metric in yaml.

I think if you want to avoid long lists of parameters in config, you can define those metrics in your class as a custom metric. Config helps to configure exactly bf-metric. That's why I prefer to add an additional level to distinguish `evaluate` and `__init__` methods.

AlexeyKozhevin:
Remove `README.md`, add `validator.rst`, add `check_api` and `check_config` methods.

Example of validator running:
```
    PascalValidator.check_api(methods=['train', 'inference'])
    val = PascalValidator('validator.yaml')
    val.check_config(keys=['train|pretrained']) # check if config has at least one key
    val.run()
    print(val.metrics)
```

docs/intro/validator.rst
```diff
@@ -0,0 +1,225 @@
...
+------------
+::
+
+    import ...
+
+    class PascalValidator(Validator):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.model_config = ...
+            self.ds = PascalSegmentation(bar=True)
```
roman-kh, line 38:
```suggestion
            self.dataset = PascalSegmentation(bar=True)
```

docs/intro/validator.rst
```diff
@@ -0,0 +1,225 @@
...
+Validator is an instrument to unify the process of model training and validation.
+There are two main reasons to wrap your model using classes from `Validator` submodule:
+
+- make the structure of the model lifecycle clear,
+- provide API to automized model training and validation.
+
+All you need is:
+
+- define class inherited from `Validator`,
+- configure it by `yaml` file,
```
roman-kh, line 16: yaml is not required. Describe all options

docs/intro/validator.rst
```diff
@@ -0,0 +1,225 @@
...
+Getting started
+===============
+Validator is an instrument to unify the process of model training and validation.
+There are two main reasons to wrap your model using classes from `Validator` submodule:
+
+- make the structure of the model lifecycle clear,
+- provide API to automized model training and validation.
+
+All you need is:
+
```
roman-kh, line 14: There are several sub-processes.
A model developer has to create a Validator subclass, but he might have nothing to do with the validation process.
While an acceptor has to define validation procedure and thus configure yaml file or even several files.

docs/intro/validator.rst
```diff
@@ -0,0 +1,225 @@
...
+- define class inherited from `Validator`,
+- configure it by `yaml` file,
+- execute the following code: ::
```
roman-kh, line 17: Let's add a default script which runs validation process

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,271 @@
...
+
+        self.train_dataset = None
+        self.from_train = None
+        self.test_dataset = None
+        self.targets = None
+        self.predictions = None
+
+        self.metrics = {}
+
+    def train_loader(self, path=None, **kwargs):
```
roman-kh, line 118: Since other methods use verb-based naming this sounds ambiguous (it loads but the name suggests that it trains).
Let's rename it to `load_train`or even `load_train_dataset`

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,271 @@
...
+        Return
+        ------
+        path : str or None
+            If function is not defined in child class, return `path`. It will be
+            used as first argument of `train`.
+         """
+        _ = kwargs
+        return path
+
+    def test_loader(self, path=None, **kwargs):
```
roman-kh, line 135: Same - `load_test`

docs/intro/validator.rst
```diff
@@ -0,0 +1,225 @@
...
+            targets = np.array(test_ppl.v('targets'))
+            predictions = np.array(test_ppl.v('predictions'))
+            return targets, predictions
+
+
+validator.yaml
+--------------
+::
+
+    train:
```
roman-kh, line 64: Why is train empty?

docs/intro/validator.rst
```diff
@@ -0,0 +1,225 @@
...
+----------------------------------------
+
+The same as `train_loader` but for `test`.
+
+`load_model(self, path=None, **kwargs)`
+---------------------------------------
+
+Loader for pretrained model. Let's make example above more complex::
+
+    train:
```
roman-kh, line 112: Why is train here if pretrained is used?

docs/intro/validator.rst
AlexeyKozhevin: `train` has one input `dataset` which is created in `load_train_dataset`. We don't need to define any parameters so we define empty `train` stage in config.

docs/intro/validator.rst
AlexeyKozhevin: "Note that when you define `pretrained` key in your config, train section will be skipped."

Just for fun

docs/intro/validator.rst
AlexeyKozhevin: I have such a script in `ml_project_template`. Do we really need it in `batchflow` if `validator` is just an interface?

docs/intro/validator.rst
AlexeyKozhevin: I've solve to change it:
```
        if 'train' in self.config and 'pretrained' in self.config:
            raise ValueError("Both 'train' and 'pretrained' was founded.")
```

roman-kh: Hmm. If we train a pretrained model, should we define both pretrained and train or load model implicitly within train?
I believe both is a better option.

AlexeyKozhevin: You describe a special case but you can implement it using train kwargs. To load and train model you can write the method `train` in the following way:
```
def train(dataset, path_to_model, **kwargs):
    model = load_model(path_to_model)
    ppl = (dataset.p
               ...)
    ...
```
In yaml you can write

```
     train:
        path_to_model: /path/to/model
```

I don't think that we need to implement that custom logic as base functional.

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+
+        if 'train' in self.config and 'pretrained' in self.config:
+            raise ValueError("Both 'train' and 'pretrained' was founded.")
+
+        self._pretrained = self.config.get('pretrained', {})
+        if 'pretrained' in self.config:
+            if self._pretrained is None:
+                self._pretrained = {}
+            elif isinstance(self._pretrained, str):
+                self._pretrained = {'path': self._pretrained}
```
SergeyTsimfer, line 89: Looks cumbersome, maybe:
`self._pretrained = {'path': config['pretrained']} if isinstance(config.get('pretrained'), str) else config.get('pretrained', {})`
split into multiple lines. Same for `self._train_ds`, `self._test_ds`

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+        elif isinstance(config, dict):
+            self.config_path = None
+            self.config = deepcopy(config)
+        elif isinstance(config, Config):
+            self.config_path = None
+            self.config = deepcopy(config.config)
+        elif config is None:
+            self.config = {}
+
+        self.config = {**self.config, **kwargs}
```
SergeyTsimfer, line 79: Maybe you can create a dictionary with default keys (pretrained, train, test) and create config like:
`self.config = {**defaults, **self.config, **kwargs}`?
That would spare you quite a few lines and checks

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+        else:
+            _warning = lambda msg: msg
+        for key in keys:
+            cond = all([_key not in self.config for _key in key.split('|')])
+            if cond:
+                error = True
+                _ = _warning('Key "{}" was not founded in config: {}'.format(key, self.config))
+        return error
+
+    def run(self):
```
SergeyTsimfer, line 296: I would rather move `run` (and `_run`/`_run_cv`) higher, as most of the other methods (like train/test/inference) are mere ~`abstract` placeholders

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+        """ Run validator """
+        if 'cv' not in self.config:
+            self._run()
+        else:
+            self._run_cv()
+
+    def _run(self):
+        loader_kwargs = self.config.get('load_train_test_dataset') or {}
```
SergeyTsimfer, line 304: A simple `logging` of actions would be very welcome

batchflow/tests/validator_test.py
```diff
@@ -0,0 +1,186 @@
...
+
+            np.random.seed(42)
+            x = np.random.random(size).reshape(-1, 1)
+            self.train_dataset = x, 2 * x
+            return self.train_dataset
+
+        def load_test_dataset(self, size=50):
+            self.stages += ['load_test_dataset']
+
+            np.random.seed(6)
```
SergeyTsimfer, line 48: .

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+                raise NotImplementedError(msg)
+        elif warning:
+            _warning = warnings.warn
+        else:
+            _warning = lambda msg: msg
+        for key in keys:
+            cond = all([_key not in self.config for _key in key.split('|')])
+            if cond:
+                error = True
+                _ = _warning('Key "{}" was not founded in config: {}'.format(key, self.config))
```
SergeyTsimfer, line 293: Was not found
There are quite a lot of strange English constructions in the docstrings, please, check it with some auto-correction tool.

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+import warnings
+import inspect
+from copy import deepcopy
+import yaml
+import numpy as np
+
+from ...pipeline import METRICS
+from ... import Config
+
+def _get_class_that_defined_method(meth):
```
SergeyTsimfer, line 12: `get_method_owner`, maybe?

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+            self._run_cv()
+
+    def _run(self):
+        loader_kwargs = self.config.get('load_train_test_dataset') or {}
+        train_dataset, test_dataset = self.load_train_test_dataset(**loader_kwargs)
+
+        if 'pretrained' in self.config:
+            self.load_model(**self._pretrained)
+        else:
+            train_dataset = train_dataset or self.load_train_dataset(**self._train_ds)
```
SergeyTsimfer, line 310: If config looks like `{'train': Dataset(10)}` this line would fail

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+        if isinstance(self._metrics, str):
+            self._metrics = [item.strip() for item in self._metrics.split(',')]
+
+        self._cv = self.config.get('cv', {})
+        if self._cv is None:
+            self._cv = {}
+        elif isinstance(self._cv, str):
+            self._cv = {'path': self._cv}
+
+        self.train_dataset = None
```
SergeyTsimfer, line 111: `self.train_dataset` and `self.test_dataset`  are not used anywhere. I suppose they are here to demonstrate the possibility to use such attributes as transport for entities; though, I would highly discourage it as it is very implicit.

batchflow/models/validator/validator.py
```diff
@@ -0,0 +1,337 @@
...
+            self.load_model(**self._pretrained)
+        else:
+            train_dataset = train_dataset or self.load_train_dataset(**self._train_ds)
+            self.train(train_dataset, **self.config['train'])
+
+        test_dataset = test_dataset or self.load_test_dataset(**self._test_ds)
+        self.targets, self.predictions = self.inference(test_dataset, **self.config['test'])
+
+        metrics_kwargs = {key: value for key, value in self.config.items() if key in self._metrics}
+        self.metrics = self.compute_metrics(self.targets, self.predictions, *self._metrics, **metrics_kwargs)
```
SergeyTsimfer, line 317: Maybe, return everything that is computed in this function, instead of silently storing it in the attributes of an instance?

batchflow/models/validator/validator.py
AlexeyKozhevin: Done

batchflow/models/validator/validator.py
AlexeyKozhevin: Done

batchflow/models/validator/validator.py
AlexeyKozhevin: I've removed them

batchflow/models/validator/validator.py
AlexeyKozhevin: Do you mean only `self.metrics`?

batchflow/models/validator/validator.py
AlexeyKozhevin: Yeap, `'train'` must be a dict. From docs:
```
`load_train_dataset(self, path=None, **kwargs)`
path and kwargs are from config <load_train_dataset>
```

PR #514, opened 2020-06-29
Add release action

### LINKED ISSUE #490: Create release gh action
Make release actions:
- build docs
- calc test coverage
- create python package and upload it to pypi.

Actions should fire at various statuses: created/edited, published, etc.

cdtn:
Closes #490.


roman-kh:
Remove coverage calculation from other actions

cdtn:
After merging this PR to `master`, workflow `release` will be triggered each time a tag matching `v*` pattern is created in repository.

What it does:
1. Builds docs and pushes them to `gh-pages` branch.
2. Calculates test coverage and uploads it to Codecov.
3. Builds the package (`batchflow`) and uploads it to `test.pypi.org` (should be changed to `pypi.org` when tested in master).

Notes:
- Outdated `actions/checkout@v1` is used in workflow. It has some drawbacks and should be replaced by `actions/checkout@v2` as soon as `git` package is updated in `ds-py3` container to the latest version.
- After `actions/checkout` is updated, the stage of pushing docs can be implemented without 3rd party action `ad-m/github-push-action`.
- A secret `PYPI_TOKEN` **must be created** in repository to allow uploads to `pypi.org`.
- Tagging commits without modifying the version defined in `batchflow/__init__.py`, will result in an error due to package upload attempt with version number already uploaded earlier.

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+    - uses: actions/checkout@v1
+
+    - name: Build docs
+      run: |
+        export LC_ALL="C"
+        pip3 install --upgrade sphinx sphinx-rtd-theme
+        git checkout --orphan gh-pages
+        rm -rf *
+        git fetch
+        git checkout origin/master -- batchflow docs
```
roman-kh, line 26: Replace last 3 lines with
```suggestion
        git reset
        git clean -fdx
        git checkout ${{ github.sha }} -- batchflow docs
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
+name: release
+
+on:
+  push:
+    tags:
+      - v*
```
roman-kh, line 6: The job will run when any tag starting with `v` is pushed (including `v1.4.1-beta2`) to any branch, which is not what we want

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+
+    runs-on: ubuntu-latest
+
+    container:
+      image: analysiscenter1/ds-py3:cpu
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Build docs
```
roman-kh, line 19: This is a duplication of `doc` workflow.

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+      uses: ad-m/github-push-action@master
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        branch: gh-pages
+        force: True
+
+    - name: Clear docs
+      run: rm -rf *
+
+    - uses: actions/checkout@v1
```
roman-kh, line 50: Why not run it as a separate job?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+        pytest -m "not slow" --cov=./ --cov-report=xml
+
+    - name: Upload coverage to Codecov
+      run: |
+        pip3 install -U codecov
+        codecov -t ${{ secrets.CODECOV_TOKEN }}
+
+    - uses: actions/checkout@v1
+
+    - name: Install dependencies
```
roman-kh, line 64: Why not run packaging as a separate job?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+      run: pip3 install --upgrade setuptools wheel twine
+
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ '__token__' }}
```
roman-kh, line 69: What is it? Why not `secrets.PYPI_USERNAME`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+        TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+      run: |
+        python3 setup.py sdist bdist_wheel
+        twine upload --repository testpypi dist/*
```
roman-kh, line 73: Not test!

roman-kh (changes requested):

.github/workflows/release.yml
cdtn: This is done on purpose. Since the workflow can only be truly tested, when merged into `master`, I thought it will be a good idea to use [test.pypi.org](test.pypi.org) first.

.github/workflows/release.yml
cdtn: Issue #490 says that release action should build docs, that's why it's here.

Since it may be a good idea too keep public docs up to date with latest public release, I suggest resolving this duplication by either total `doc` workflow removal, or by forcing it run only on releases (and remove building docs from this workflow).

.github/workflows/release.yml
cdtn: We can either define a pattern for tags of "solid" releases, that should be uploaded to PyPi, or additionally check if the tagged commit is in release branch (e.g. `master`). I suggest the first option, since Github Actions doesn't currently allow combining tag and branch checks in a convenient way. How about `v[0-9]\.[0-9]+\.[0-9]+`?

roman-kh: Use `release` [event](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release)

.github/workflows/release.yml
roman-kh: The doc should definitely be build and published during release, but `release` is not the only trigger

.github/workflows/release.yml
roman-kh: There is [job conditioning](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds)

.github/workflows/release.yml
roman-kh: See `job.needs`

.github/workflows/release.yml
roman-kh: Testing should be done before publishing a release

.github/workflows/release.yml
roman-kh: What is `${{'__token__'}}`?

cdtn: Just a string constant, that tells PyPi API to expect token instead of password.

roman-kh: Where does it take its value from?

cdtn: From password field on the following line:
`TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}`

PR #515, opened 2020-06-30
unit tests for Batch._assemble fixes #241

### LINKED ISSUE #241: Make tests for _assemble

a-arefina:
test cases for `dst_default` checking might be redundant, because it seems, that we actually can rid of using `deafault_dst` in assemble


roman-kh:
Fix the conflict please

batchflow/tests/assemble_test.py
```diff
@@ -0,0 +1,45 @@
...
+    # dst_default checking
+    ([0, 1],           dict(src='c1'),                 dict(c1=[0, 1], c2=None)),
+    ([[0, 2], [1, 3]], dict(dst_default='components'), dict(c1=[0, 1], c2=[2, 3])),
+])
+def test_assemble(all_results, kwargs, res):
+    """ ensure that values from `all_results` are put into components properly """
+    batch = FakeBatch(np.arange(2))
+    batch._assemble(np.asarray(all_results), **kwargs)
+
+    def check_res(item, true_val):
```
roman-kh, line 38: Why is it within the test?

a-arefina: Because it is not needed elsewhere. Comparisons of arrays in other tests don't have the `if true_val is None` branch

roman-kh: Nested functions have performance penalty which is not severe in this particular case but is still unnecessary.

PR #519, opened 2020-09-03
Fix Notifier description

SergeyTsimfer:
More types for valid description in `Notifier`


batchflow/notifier.py
```diff
@@ -337,7 +337,7 @@ def create_description(self, iteration):
             name = container['name']
             if isinstance(source, (str, NamedExpression)):
                 value = container['data'][iteration]
```
roman-kh, line 339: `value = getattr(value, 'tolist', lambda: value)()`
this will convert value to pure python scalar

PR #522, opened 2020-09-11
Fix pylint

nikita-klsh:
This PR fixes pylint output in the latest docker image analysiscenter1/ds-py3. 

A lot of `missing-function-docstring` and `abstract-method` warnings happened in `batchflow/models/torch` folder we cant fix, because they occur on the torch side. They are fixing [this](https://github.com/pytorch/pytorch/issues/42305) and [this](https://github.com/pytorch/pytorch/issues/43057). 

So we can:
1.  Close eyes on this warnings.
2.  Temporary set additional restrictions for pylint for `torch` folder only like its currently done. 

@roman-kh

### DIFF
--- a/batchflow/dataset.py
+++ b/batchflow/dataset.py
@@ -75,6 +75,7 @@ class Dataset(Baseset):
         self._attrs = None
         kwargs['_copy'] = kwargs.get('_copy', copy)
         self.n_splits = None
+        self.train, self.test = None, None
 
         cv_kwargs = {item: kwargs.pop(item) for item in ['method', 'n_splits', 'shuffle'] if item in kwargs}
         if cv_kwargs.get('n_splits') is not None:
--- a/batchflow/dataset.py
+++ b/batchflow/dataset.py
@@ -75,7 +75,6 @@ class Dataset(Baseset):
         self._attrs = None
         kwargs['_copy'] = kwargs.get('_copy', copy)
         self.n_splits = None
-        self.train, self.test = None, None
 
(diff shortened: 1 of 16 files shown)


roman-kh:
@nikita-klsh It seems `pytorch` issues are solved already.

nikita-klsh:
> @nikita-klsh It seems `pytorch` issues are solved already.

@roman-kh There are no releases with updates yet. Or we want to build it from github?

nikita-klsh:
Running `pylint` with `torch 1.7.0` still raises warnings about `forward` methods missing docs. 
Torch solved the issue and now the method formally obtains the doc, i.e. `nn.Module.forward.__doc__ is not None`, even tho docs are not explicitly speсified in the method's body. 
I suppose this might be `pylint` false warning.

batchflow/dataset.py
```diff
@@ -75,7 +75,6 @@ def __init__(self, index, batch_class=Batch, *args, preloaded=None, cast_to_arra
         self._attrs = None
         kwargs['_copy'] = kwargs.get('_copy', copy)
         self.n_splits = None
-        self.train, self.test = None, None
```
roman-kh: Why do you think this is a better approach?

nikita-klsh: Looks weird to reinitialize attributes, that have already been initialized in the parent class `Baseset`

nikita-klsh: I return it as it was initially and plug `pylint: disable=attribute-defined-outside-init` stub.
In the code we assign values to `self.train / self.test / self.validation` attributes that WERE defined in the `init`.
It looks like pylint false positive

PR #524, opened 2020-09-21
Default timeout for Research is None; RC can be without name

AlexeyKozhevin:
The previous default value for timeout (5 minutes) in Research doesn't allow to execute massive tasks in callables.
Now by default timeout is None.

RC (ResearchConfig) named expression can be used without the specification of the executable unit.

### DIFF
--- a/batchflow/research/named_expr.py
+++ b/batchflow/research/named_expr.py
@@ -24,12 +24,11 @@ class REU(ResearchNamedExpression): # ResearchExecutableUnit
             _experiment = experiment
         else:
             _experiment = [experiment]
-        if self.name is not None:
-            res = [item[self.name] for item in _experiment]
-            if len(_experiment) == 1:
-                return res[0]
-            return res
-        return experiment
+        name = self.name if self.name is not None else list(_experiment[0].keys())[0]
+        res = [item[name] for item in _experiment]
+        if len(_experiment) == 1:
+            return res[0]
+        return res
 
 class RP(REU): # ResearchPipeline
     """ NamedExpression for Pipeline in Research """
@@ -59,7 +58,7 @@ class RI(ResearchNamedExpression): # ResearchIteration
 
 class RC(REU): # ResearchConfig
     """ NamedExpression for Config of the ExecutableUnit """
-    def __init__(self, name):
+    def __init__(self, name=None):
         super().__init__(name=name)
 
     def get(self, **kwargs):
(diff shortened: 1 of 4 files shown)


batchflow/research/workers.py
```diff
@@ -107,14 +107,15 @@ def __call__(self, queue, results):
...
-                            if signal is None and (time.time() - last_update_time.value) / 60 > self.timeout:
-                                p = psutil.Process(pid)
-                                p.terminate()
-                                message = 'Job {} [{}] failed in {}'.format(job[0], pid, self.worker_name)
-                                self.logger.info(message)
-                                final_signal.exception = TimeoutError(message)
-                                results.put(copy(final_signal))
-                                break
+                            if signal is None:
+                                if self.timeout is not None and (time.time() - last_update_time.value) / 60 > self.timeout:
```
a-arefina, line 111: it is better to unite these 2 conditions into one

batchflow/research/named_expr.py
```diff
@@ -24,12 +24,11 @@ def get(self, **kwargs):
             _experiment = experiment
         else:
             _experiment = [experiment]
-        if self.name is not None:
-            res = [item[self.name] for item in _experiment]
-            if len(_experiment) == 1:
-                return res[0]
-            return res
-        return experiment
+        name = self.name if self.name is not None else list(_experiment[0].keys())[0]
```
a-arefina, line 27: This way of getting a missing name seems rather obscure. Is the first key in the first experiment that is passed to eval_expr is always the name we want? Or they are all equal?

AlexeyKozhevin: Actually, all configs are the same for all units

batchflow/research/workers.py
AlexeyKozhevin: I would like to keep them separate because there are condition branches that only deal with signal conditions.

a-arefina (commented):
It would be great if you write some tests, at least for `RC`

batchflow/research/workers.py
```diff
@@ -107,14 +107,15 @@ def __call__(self, queue, results):
...
+                                    p = psutil.Process(pid)
+                                    p.terminate()
+                                    message = 'Job {} [{}] failed in {}'.format(job[0], pid, self.worker_name)
```
SergeyTsimfer, line 114: Please, change the message to something more descriptive like `failed because of timeout`

a-arefina (approved):

PR #527, opened 2020-10-14
Add notebook utils

SergeyTsimfer:
This PR adds:

- function to get the absolute path of current Jupyter Notebook, for example, `/path/path/a/b/c/My_notebook.ipynb`

- function to get the name of current Jupyter Notebook, for example, `My_notebook`

- function to run linter on the notebook and produce a custom report with results. Function allows to provide additional options like `rcfile` and disable some of the errors. Example of the output:

image


batchflow/utils.py
```diff
@@ -8,6 +11,12 @@
 from matplotlib import pyplot as plt
 from matplotlib import colors as mcolors
 
+import ipykernel
+import requests
+from notebook.notebookapp import list_running_servers
+from nbconvert import PythonExporter
+from pylint import epylint as lint
```
roman-kh, line 18: Add them all to requirements.txt

batchflow/utils.py
```diff
@@ -371,3 +380,147 @@ def save_data_to(what, where, **kwargs):
...
         else:
             where[i] = item
+
+
+def get_notebook_path():
+    """ Return the full absolute path of the current jupyter notebook,
+    for example, `/path/path/path/My_notebook_title.ipynb`.
+    """
+    kernel_id = re.search('kernel-(.*).json',
+                          ipykernel.connect.get_connection_file()).group(1)
```
a-arefina, line 390: What will kernel_id be, if this is run not from a notebook? I'd add a way to leave this function without iterating servers

batchflow/utils.py
```diff
@@ -371,3 +380,147 @@ def save_data_to(what, where, **kwargs):
...
+    path : str, optional
+        Path to run linter on. If not provided, the callee notebook is linted.
+    options : str
+        Additional flags for linter execution, for example, the pylint configuration options.
+    printer : callable
+        Method for displaying results.
+    ignore_comments : bool
+        Whether to ignore markdown cells and comments in code.
+    ignore_codes : sequence
+        Pylint errors to ignore. By default, invalid names and import errors are disabled.
```
a-arefina, line 425: I would list exact pylint error names

batchflow/utils.py
```diff
@@ -371,3 +380,147 @@ def save_data_to(what, where, **kwargs):
...
+    keep_script : bool
+        Whether to keep temporal `.py` file after command execution.
+    return_report : bool
+        If True, then this function returns the string representation of produced report.
+        If False, then 0 is returned.
```
a-arefina, line 430: Why not None?

batchflow/utils.py
```diff
@@ -371,3 +380,147 @@ def save_data_to(what, where, **kwargs):
...
+
+        if 'rated' in error_line:
+            report_.insert(0, error_line.strip(' '))
+            report_.insert(1, '-' * (len(error_line) - 1))
+            report_.insert(2, '')
+
+    printer('\n'.join(report_))
+
+    # Cleanup
+    if keep_script is False:
```
a-arefina, line 521: I would change to 'if not keep_script'

a-arefina (commented):
It is a great idea to implement these utilities.

Please consider the case when they are run not from a notebook. They should terminate graciously without extra work

batchflow/utils.py
```diff
@@ -382,10 +382,27 @@ def save_data_to(what, where, **kwargs):
...
 
 
+def in_notebook():
+    """ Return True if in Jupyter notebook and False otherwise. """
+    try:
+        shell = get_ipython().__class__.__name__
+        if shell == 'ZMQInteractiveShell':
+            return True
+        if shell == 'TerminalInteractiveShell':
+            return False
```
a-arefina, line 392: I seems that you don't need this branch :) You return False anyway

a-arefina (commented):
One tiny comment

batchflow/utils.py
roman-kh, line 18: This will required a lot of packages to be pre-installed for batchflow even if you don't use these functions.
Make them optional.

roman-kh (changes requested):

batchflow/utils.py
SergeyTsimfer: I can either use `try/except` block on top of the module, or move those imports to the functions that use them. What is the preferred way of making modules optional?

roman-kh: Top imports are much more explicit, thus better.
But local imports are also acceptable if they are mentioned at the top.

roman-kh: I would also suggest to split `utils` into separate modules for plotting, notebooks, etc

cdtn (approved):

PR #533, opened 2020-10-28
GPU monitor

SergeyTsimfer:
As it turned out, our previous `GPUMemory` monitor was measuring memory utilization: that is, how often do we push to it/write from it. New implementation allows to look at the proportion of memory taken on GPU.

Also, this PR adds a wrapper around `list` to hold multiple monitors, returned by `monitor_resource` context manager: this allows to easily visualize them on one figure:

```
with monitor_resource(['memory', 'uss', 'gpu', 'gpu_memory', 'gpu_memory_utilization']) as monitors:
    # intensive computations

monitors.visualize() # note that otherwise, this variable is a completely valid list
```


batchflow/monitor.py
```diff
@@ -215,12 +271,14 @@ def get_usage(gpu_list=None, **kwargs):
...
+    GPUMemoryUtilizationMonitor: ['gpu_memory_utilization']
 }
 
 MONITOR_ALIASES = {alias: monitor for monitor, aliases in MONITOR_ALIASES.items()
                    for alias in aliases}
 
 
+
 @contextmanager
 def monitor_resource(resource='memory', frequency=0.5, **kwargs):
```
roman-kh, line 283:
```suggestion
def resource_monitor(resource='memory', frequency=0.5, **kwargs):
```

SergeyTsimfer: The way it is now, it is an action: `while monitoring those resources, do...:`

PR #535, opened 2020-11-02
Torch improvements

SergeyTsimfer:
This PR removes the `train_steps` construction from `TorchModel`. Along the way, I've changed inner storages for iteration info and sync counters: now they are all integers and zero-based, which makes them easier to understand.

Also, I changed the `'.'` operation letter for concatenation to `'|'`, as dots are not allowed in the module name by `Pytorch` convention.

Moreover, I've made sure that all the tests are working: one of them was failing due to the new (one-year-old) syntax of `decays`.

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -267,8 +267,10 @@ class TorchModel(BaseModel, VisualizationMixin):
     PRESERVE = [
         'full_config', 'config', 'model',
         'input_names', 'input_shapes', 'target_shape', 'classes',
-        'loss', 'optimizer', 'decay',
-        'sync_counter', 'microbatch'
+        'loss', 'optimizer', 'decay', 'decay_step',
+        'sync_counter', 'microbatch',
+        'iteration', 'iter_info', 'lr_list', 'syncs', 'decay_iters',
+        '_loss_list', 'loss_list'
     ]
 
     def __init__(self, config=None):
@@ -304,6 +306,8 @@ class TorchModel(BaseModel, VisualizationMixin):
         self.lr_list = []
         self.syncs = []
         self.decay_iters = []
+        self._loss_list = []
+        self.loss_list = []
 
         # Profile kernels used
         self.profilers = []
@@ -872,6 +876,9 @@ class TorchModel(BaseModel, VisualizationMixin):
             output = self._train(*_inputs, _targets, fetches=fetches, sync_frequency=sync_frequency*steps)
             outputs.append(output)
 
+        # Store the average value of loss over the entire batch
+        self.loss_list.append(np.mean(self._loss_list[-steps:]))
+
         if use_lock:
             self.model_lock.release()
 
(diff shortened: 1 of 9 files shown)


batchflow/models/torch/blocks.py
```diff
@@ -207,7 +207,7 @@ def __init__(self, inputs=None, layout='nacd', filters=None, kernel_size=3, stri
             strides = [1, strides]
             filters = [growth_rate * bottleneck, filters]
 
-        layout = 'R' + layout + '.'
+        layout = 'R' + layout + '|'
```
roman-kh, line 210: Change `tf` as well then

batchflow/models/torch/base.py
```diff
@@ -916,84 +891,82 @@ def train(self, *args, feed_dict=None, fetches=None, use_lock=True, train_mode='
...
+
+        # Whether to update weights or keep accumulating
+        if self.sync_counter == sync_frequency - 1:
+            # Correct accumulated gradients
+            for p in self.model.parameters():
+                if p.grad is not None:
+                    p.grad /= sync_frequency
+
+            # Store learning rate: once per sync
+            # Note: we do it before decay, so it is actual used LR on this iteration
```
a-arefina, line 930:
```suggestion
            # Note: we do it before decay, so it is actual LR used on this iteration
```

batchflow/models/torch/base.py
```diff
@@ -916,84 +891,82 @@ def train(self, *args, feed_dict=None, fetches=None, use_lock=True, train_mode='
...
-
-                if decays:
-                    for decay, decay_step in zip(decays, decay_steps):
-                        step_condition = int(1 + step['iter'] - decay_step['first_iter']) % decay_step['frequency'] == 0
-                        range_condition = decay_step['first_iter'] <= step['iter'] <= decay_step['last_iter']
-                        if step_condition and range_condition:
-                            decay.step()
+        # Apply model, compute loss and gradients
+        predictions = self.model(inputs)
+        loss = sum(loss_fn(predictions, targets) for loss_fn in self.loss) / len(self.loss)
```
a-arefina, line 916: same expression is used in `predict` - whdt about extracting a method?

batchflow/models/torch/base.py
```diff
@@ -916,84 +891,82 @@ def train(self, *args, feed_dict=None, fetches=None, use_lock=True, train_mode='
...
+        # Store outputs
+        output_container = {
+            'predictions': predictions,
+            'loss': loss,
+        }
 
         config = self.full_config
-        additional_outputs = self.output(inputs=predictions, predictions=config['predictions'],
+        additional_outputs = self.output(inputs=predictions,
+                                         predictions=config['predictions'],
```
a-arefina, line 962: this naming is rather unclear, but I have no idea how to fix it

batchflow/models/torch/base.py
SergeyTsimfer: I thought about splitting the procedure of training into multiple blocks so you could configure each of them in your inherited class:

- `compute_predictions`

- `compute_loss`

- `optimizer_step`

etc. I suggest that we do it in separate PR, as it would make the diff in this one unreadable

PR #536, opened 2020-11-03
Efficientnet

a-arefina:
EfficientNet for torch


batchflow/models/torch/blocks.py
```diff
@@ -158,6 +158,89 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
...
+            if use_res_connect:
+                block_params['layout'] += 'R'
+
+            if expand_ratio != 1:
+                block_params['layout'] += 'cna'
+                block_params['filters'].append(inner_filters)
+                block_params['kernel_size'].append(1)
+                block_params['strides'].append(1)
+
+            block_params['layout'] += 'wna'
```
SergeyTsimfer, line 216: Make this layout a parameter

batchflow/models/torch/blocks.py
```diff
@@ -158,6 +158,89 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
...
+            block_params['filters'].append(inner_filters)
+            block_params['kernel_size'].append(kernel_size)
+            block_params['strides'].append(strides)
+
+            if se_block:
```
SergeyTsimfer, line 221: I believe that it is better to use the same convention, as in `ResBlock`: parameter `attention` that can be either `None`/`False`/`se` etc

batchflow/models/torch/blocks.py
```diff
@@ -158,6 +158,89 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
                          **kwargs)
 
 
+class MBConvBlock(ConvBlock):
```
SergeyTsimfer, line 161: Add those blocks to __init__

batchflow/models/torch/blocks.py
```diff
@@ -158,6 +158,89 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
...
+        inp_filters = get_num_channels(inputs)
+
+        layer_params = []
+        for k in range(n_reps):
+            if k > 0:
+                strides = 1
+
+            inner_filters = inp_filters * expand_ratio
+
+            use_res_connect = (strides == 1) and (inp_filters == filters)
```
SergeyTsimfer, line 203: `skip`, maybe?

batchflow/models/torch/efficientnet.py
```diff
@@ -0,0 +1,227 @@
...
+"""
+
+import torch
+
+#pylint: disable=too-many-ancestors
+from .encoder_decoder import Encoder
+from .blocks import MBConvBlock
+
+
+def swish(x):
```
SergeyTsimfer, line 15: I believe that this implementation is very inEfficient (pun intended)

batchflow/models/torch/efficientnet.py
```diff
@@ -0,0 +1,227 @@
...
+        d_factor = config.get('common/depth_factor')
+
+        for path in ('initial_block/', 'body/encoder/blocks/', 'head/'):
+            scalable = config.get(path + 'scalable')
+            if scalable:
+                for param, factor in [('filters', w_factor), ('n_reps', d_factor)]:
+                    if factor != 1:
+                        val = config.get(path + param)
+                        if val:
+                            if isinstance(val, int):
```
SergeyTsimfer, line 100: `int()` does the `floor` operation. Is that intended?

batchflow/models/torch/efficientnet.py
```diff
@@ -0,0 +1,227 @@
...
+    """ EfficientNetB1 """
+
+    resolution = 240
+
+    @classmethod
+    def default_config(cls):
+        config = super().default_config()
+
+        config['common/width_factor'] = 1.0
+        config['common/depth_factor'] = 1.1
```
SergeyTsimfer, line 121: The biggest value of `n_reps` in the config of `B0` is 4. After this scaling, we obtain `int(4 * 1.1) = 4`, so there is no actual change in the network

batchflow/models/torch/efficientnet.py
```diff
@@ -0,0 +1,227 @@
...
+    """ EfficientNetB2 """
+
+    resolution = 260
+
+    @classmethod
+    def default_config(cls):
+        config = super().default_config()
+
+        config['common/width_factor'] = 1.1
+        config['common/depth_factor'] = 1.2
```
SergeyTsimfer, line 138: Actually, still no change in `n_reps`

SergeyTsimfer (changes requested):

batchflow/models/torch/efficientnet.py
a-arefina: I'm afraid, I cannot quite understand what you mean.

batchflow/models/torch/blocks.py
a-arefina: the word 'skip' has really so many meanings. Since this is an internal variable I prefer a lager but more exact name

batchflow/models/torch/efficientnet.py
a-arefina: changed to `ceil`

batchflow/models/torch/efficientnet.py
a-arefina: fixed by using ceil in rounding

batchflow/models/torch/efficientnet.py
a-arefina: after changing `int` to `ceil` in rounding statements the issue with number of filters and repetitions being equal for B1 and B2 still exists. 

I use a simplified version of filter_rounding compared to [original code](https://github.com/tensorflow/tpu/blob/4489103eee24cd612ea3c41958c9320c45080c6f/models/official/efficientnet/efficientnet_model.py#L106) but filters and repetitions in B1 and B2 are equal when using the original `round filters` as well

So, these are the details of the original paper: B1 and B2 differ only in resolution and dropout rate

As for using simplified `round_filters` instead of the original function my point was the following. It seems over complicated, and intended for width modifiers < 1, that is reducing filters numbers. Most of extra parameters are not modified anywhere. When calculating filters with all default parameters from the original implementation one gets values that are sometimes bigger and sometimes smaller by same relatively small numbers compared to simple `int(ceil(filters*modifies))` implementation. So I decided to use this simplified version

batchflow/models/torch/efficientnet.py
SergeyTsimfer: In order to compute this activation implemented this way, `Pytorch` stores a lot of intermediate activations. The usual way to avoid this is to fuse the whole operation into one by using `Torchscript` and custom forward/backward passes. Nevertheless, it has already been implemented in the `Pytorch`: [SiLU](https://pytorch.org/docs/stable/generated/torch.nn.SiLU.html), and you can just use `'common/activation' : 'silu'` in the model configuration (if the version of `torch` package is modern enough)

a-arefina: `SiLU` is introduced in torch 1.7 (which is the most recent as for now). Are we going to stick to this version in batchflow?

SergeyTsimfer (approved):

batchflow/models/torch/blocks.py
```diff
@@ -158,6 +158,90 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
...
+        number of filters in the previous tensor. Note the `eval` usage under the hood.
+        If int, then number of filters in the output tensor. Default value is 'same'.
+    kernel_size : int
+        Convolution kernel size. Default is 3.
+    strides : int
+        Convolution stride. Default is 1.
+    attention : None, bool or str
+        If None or False, then nothing is added. Default is False.
+        If True, then add a squeeze-and-excitation block.
+        If str, then any of allowed self-attentions.
```
a-arefina, line 183:
```suggestion
        If str, then any of allowed self-attentions.
```

batchflow/models/torch/efficientnet.py
```diff
@@ -0,0 +1,229 @@
...
+                                              filters=[16, 24, 40, 80, 112, 192, 320],
+                                              expand_ratio=[1, 6, 6, 6, 6, 6, 6],
+                                              attention='se'
+                                              )
+
+        config['head'] += dict(scalable=True, layout='cna V df', kernel_size=1, strides=1, filters=1280,
+                               dropout_rate=0.2)
+
+        config['common'] += dict(activation=swish, width_factor=1.0, depth_factor=1.0)
+
```
a-arefina, line 83:
```suggestion

```

batchflow/models/torch/blocks.py
```diff
@@ -158,6 +158,90 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
...
+
+            use_res_connect = (strides == 1) and (inp_filters == filters)
+
+            block_params = dict(layout='', filters=[], kernel_size=[], strides=[])
+
+            if use_res_connect:
+                block_params['layout'] += 'R'
+
+            if expand_ratio != 1:
+                block_params['layout'] += layout
```
SergeyTsimfer, line 213: As I recall, here was the `cna` bottleneck block, and the `layout` param should be used later

batchflow/models/torch/blocks.py
```diff
@@ -158,6 +158,90 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
...
+            block_params['layout'] += 'wna'
+            block_params['filters'].append(inner_filters)
+            block_params['kernel_size'].append(kernel_size)
+            block_params['strides'].append(strides)
+
+            if attention:
+                block_params['layout'] += 'S'
+                block_params['attention'] = attention
+
+            block_params['layout'] += 'cn'
```
SergeyTsimfer, line 227: I am feeling uneasy about this layout: if the skip connection is on, we have a `Rcna wna cn+` block, and two such blocks make for a `R cna wna cn+ R cna wna cn +`, passing the data via skips with no activations at all.

I know, that it is the proposed in paper architecture; just want some intuition on why is that better than `R cna wna cn+a`, as in ResBlocks

SergeyTsimfer (changes requested):

roman-kh:
Fix the conflict please

SergeyTsimfer (approved):

PR #538, opened 2020-11-12
fix unet channels after upsampling

a-arefina:
channels in original UNet are halved after upsampling

### DIFF
--- a/batchflow/models/torch/unet.py
+++ b/batchflow/models/torch/unet.py
@@ -75,6 +75,7 @@ class UNet(EncoderDecoder):
 
         config['body/decoder/order'] = ['upsampling', 'combine', 'block']
         config['body/decoder/blocks'] += dict(layout='cna cna', kernel_size=3, filters=[512, 256, 128, 64])
+        config['body/decoder/upsample/filters'] = [512, 256, 128, 64]
 
         config['loss'] = 'ce'
         return config
@@ -92,6 +93,7 @@ class UNet(EncoderDecoder):
             config['body/embedding/filters'] = encoder_filters[-1] * 2
             config['body/decoder/num_stages'] = num_stages
             config['body/decoder/blocks/filters'] = encoder_filters[::-1]
+            config['body/decoder/upsample/filters'] = encoder_filters[::-1]
 
         return config
 
--- a/batchflow/models/torch/unet.py
+++ b/batchflow/models/torch/unet.py
@@ -75,7 +75,6 @@ class UNet(EncoderDecoder):
 
         config['body/decoder/order'] = ['upsampling', 'combine', 'block']
         config['body/decoder/blocks'] += dict(layout='cna cna', kernel_size=3, filters=[512, 256, 128, 64])
-        config['body/decoder/upsample/filters'] = [512, 256, 128, 64]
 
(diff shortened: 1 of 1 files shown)


batchflow/models/torch/unet.py
```diff
@@ -75,6 +75,7 @@ def default_config(cls):
 
         config['body/decoder/order'] = ['upsampling', 'combine', 'block']
         config['body/decoder/blocks'] += dict(layout='cna cna', kernel_size=3, filters=[512, 256, 128, 64])
+        config['body/decoder/upsample/filters'] = [512, 256, 128, 64]
```
roman-kh, line 78: Why not 
```suggestion
        config['body/decoder/upsample/filters'] = config['body/decoder/blocks/filters']
```

a-arefina: It's shorter, and the number of filters for the default configuration is not going to be changed, so there's no need to tie it to block's filters

a-arefina: I now see, that setting upsample/filters should indeed be done as you propose, but in build_config, not here

a-arefina:
> On one hand, making the architecture look exactly like described in the paper is a good thing.
> On the other, this will 100% break some of the existing code: all the configs that use this class and set custom values of `decoder/blocks/filters` and not `decoder/upsample/filters` would be changed.
> In any case, I think we can split implemented architectures into two groups: one with the exact parameters, as in the reference paper (lets for now call it `PaperUNet`), and one with blank values, so you would need to (explicitly!) set them in your config (simply `UNet`).
> This way, if you really don't want to bother, you can just use the version with predefined filters, or be as explicit as it gets and set all the values yourself.
> @a-arefina @roman-kh thoughts?

I don't like the idea of having a special `PaperUnet` that differs from the old `UNet` only in the number of upsampling filters, but if the customizable `UNet` has all blank values, it can break backward compatibility as well... 

What about throwing a warning in numpy's this-will-change-in-future-style in build config if upsampling/filters are not set?

SergeyTsimfer:
> > On one hand, making the architecture look exactly like described in the paper is a good thing.
> > On the other, this will 100% break some of the existing code: all the configs that use this class and set custom values of `decoder/blocks/filters` and not `decoder/upsample/filters` would be changed.
> > In any case, I think we can split implemented architectures into two groups: one with the exact parameters, as in the reference paper (lets for now call it `PaperUNet`), and one with blank values, so you would need to (explicitly!) set them in your config (simply `UNet`).
> > This way, if you really don't want to bother, you can just use the version with predefined filters, or be as explicit as it gets and set all the values yourself.
> > @a-arefina @roman-kh thoughts?
> 
> I don't like the idea of having a special `PaperUnet` that differs from the old `UNet` only in the number of upsampling filters, but if the customizable `UNet` has all blank values, it can break backward compatibility as well...
> 
> What about throwing a warning in numpy's this-will-change-in-future-style in build config if upsampling/filters are not set?

The new `UNet` would definitely break the existing models, if they have some unset filters; this break will be explicit and clearly visible.
The current approach applies changes to the models without you even noticing. That is what I am trying to solve with the proposal

roman-kh:
Warning is a good idea

PR #539, opened 2020-11-13
Sampler classes

akoryagin:
## Main changes

* Remove inner functions from `Sampler.sample` for `multiprocessing` (`pickle`, really) to work with `Sampler`-objects. Now all operations on samplers (`&`, `|`,..., `+`, `-`, ..., `%`) are implemented in corresponding subclasses of `Sampler`.
* Add multiprocessing-example into the `Sampler`-tutorial.


batchflow/sampler.py
```diff
@@ -266,63 +198,174 @@ def truncate(self, high=None, low=None, expr=None, prob=0.5):
...
+
+
+class OrSampler(Sampler):
+    def __init__(self, left, right, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.left = left
+        self.right = right
+
+        # calculate probs of samplers in mixture
+        _ws = np.array([self.left.weight, self.right.weight])
```
SergeyTsimfer, line 211: I think that there are a lot of unnecessary underscores in variable names; can you remove them while you are at it?

batchflow/sampler.py
```diff
@@ -266,63 +198,174 @@ def truncate(self, high=None, low=None, expr=None, prob=0.5):
...
-        result.sample = truncated
-
-        return result
+        return TruncateSampler(self, high, low, expr, prob)
+
+
+class OrSampler(Sampler):
+    def __init__(self, left, right, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.left = left
```
SergeyTsimfer, line 207: `left` and `right can be confusing
Maybe, use `bases` attribute in every derived class to hold its, well, bases?

batchflow/sampler.py
```diff
@@ -266,63 +198,174 @@ def truncate(self, high=None, low=None, expr=None, prob=0.5):
...
+        self.right = right
+
+        # calculate probs of samplers in mixture
+        _ws = np.array([self.left.weight, self.right.weight])
+        self.weight = np.sum(_ws)
+        self._normed = _ws / np.sum(_ws)
+
+    def sample(self, size):
+        """ Sampling procedure of a mixture of two samplers.
```
SergeyTsimfer, line 216: This is not an informative description: readers already know that this is a class to create mixture of samplers. A documentation of what is actually going on would help

batchflow/sampler.py
```diff
@@ -266,63 +198,174 @@ def truncate(self, high=None, low=None, expr=None, prob=0.5):
...
+        self.sampler = sampler
+        self.high = high
+        self.low = low
+        self.expr = expr
+        self.prob = prob
+
+    def sample(self, size):
+        """ Sampling method of a sampler subjugated to truncation.
+        """
+        if size == 0:
```
SergeyTsimfer, line 264: Why is that needed here, and not in any of the other `Sampler`-derived classes?

batchflow/sampler.py
```diff
@@ -266,63 +198,174 @@ def truncate(self, high=None, low=None, expr=None, prob=0.5):
...
+                    cond &= np.less_equal(expr(sample).reshape(batch_size, -1), high).all(axis=1)
+                else:
+                    cond &= np.less_equal(sample, high).all(axis=1)
+
+            if high is None and low is None:
+                cond &= expr(sample).all(axis=1)
+
+            # check that truncation-prob is not to small
+            _share = np.sum(cond) / batch_size
+            if _share < SMALL_SHARE and ctr > 0:
```
SergeyTsimfer, line 298: Why would we need this? What is the situation where you actually want to get this error?
I mean, in most cases, you just want your generated points: it can take you 0.01ms or 1ms, if you needed to resample 100 times, but rarely do you care for such small margins. You just want your points.
Can we remove this or, at least, make a `I don't care how much time it takes just sample the points` flag, which is set to `True` by default?

batchflow/sampler.py
```diff
@@ -266,63 +198,174 @@ def truncate(self, high=None, low=None, expr=None, prob=0.5):
...
+
+            # get points from region of interest
+            samples.append(sample[cond])
+            cumulated += np.sum(cond)
+            ctr += 1
+
+        return np.concatenate(samples)[:size]
+
+
+class BaseOperationSampler(Sampler):
```
SergeyTsimfer, line 309: You can move it to `arithmetize` decorator, as well as all those two-lines classes

batchflow/sampler.py
```diff
@@ -266,63 +198,174 @@ def truncate(self, high=None, low=None, expr=None, prob=0.5):
...
+class RPowSampler(BaseOperationSampler):
+    operation = '__rpow__'
+
+class RFloordivSampler(BaseOperationSampler):
+    operation = '__rfloordiv__'
+
+class RModSampler(BaseOperationSampler):
+    operation = '__rmod__'
+
+classes = dict(zip(['__add__', '__mul__', '__truediv__', '__sub__', '__pow__', '__floordiv__', '__mod__',
```
SergeyTsimfer, line 365: This should be at the top of the module and upper-cased

SergeyTsimfer (changes requested):
Also, why some of the variables like `SMALL_SHARE` and `ALIASES` are not class-attributes of `Sampler`?

batchflow/sampler.py
SergeyTsimfer: If anything, it should be a warning: `Resampling points 1232131 times, so make sure you know what you are doing`

batchflow/sampler.py
akoryagin: I don't believe I can. If I do it like that (or in a similar manner):
```
def arithmetize(cls):
    class BaseOperationSampler(cls):
        ...
    class AddSampler(BaseOperationSampler):
        operation = '__add__'
        ...
    classes = ...
    for oper in ['__add__', ...]:
        def transform(self, other, fake=oper):
            _class = classes[fake]
            return _class(self, other)
        setattr(cls, oper, transform)

    return cls
```
`Sampler`-instances won't be _picklable_ anymore, as the classes are defined in a local namespace of `arithmetize`.
`Pool` will crash with `AttributeError: Can't pickle local object 'arithmetize.<locals>.AddSampler'`.

Check it out [here](http://82.148.26.146:8888/lab?) if interested.

batchflow/sampler.py
akoryagin: Again, to my understanding, this variable can be defined only after all those two-line classes. Which can be defined only after original `Sampler`.

batchflow/sampler.py
SergeyTsimfer: I believe that it is _possible_ by adding the names of dynamically generated classes to the module-scope, or by adding the `__reduce__` method to all of those classes; nevertheless, that is too much trouble

roman-kh:
Have you agreed ob everything? Merging?

akoryagin:
> Have you agreed ob everything? Merging?

Nope, not yet

akoryagin:
> > Have you agreed ob everything? Merging?
> 
> Nope, not yet

All fixed, @SergeyTsimfer @roman-kh ?

batchflow/sampler.py
```diff
@@ -45,13 +46,12 @@ def _get_method_by_alias(alias, module, tf_distributions=None):
 
     return fullname
 
+
```
SergeyTsimfer, line 49: Why `ALIASES` and `_get_method_by_alias` are not part of the base `Sampler` class?

SergeyTsimfer: As well as the `MAX_ITERS` constant

SergeyTsimfer (approved):

PR #540, opened 2020-11-18
Fix load model

AlexeyKozhevin:
Previously, `load_model` was executed each iteration. Now only once.

### DIFF
--- a/batchflow/pipeline.py
+++ b/batchflow/pipeline.py
@@ -1080,12 +1080,8 @@ class Pipeline:
         args, kwargs
             model-specific parameters (like paths, formats, etc)
         """
-        if mode == 'static':
-            self.models.load_model(mode, name, model_class, *args, **kwargs)
-            return self
-        return self._add_action(LOAD_MODEL_ID, *args,
-                                _args=dict(mode=mode, model_class=model_class, model_name=name),
-                                **kwargs)
+        self.before.load_model(mode, name, model_class, *args, **kwargs)
+        return self
 
     def _exec_load_model(self, batch, action):
         mode = self._eval_expr(action['mode'], batch=batch)
--- a/batchflow/pipeline.py
+++ b/batchflow/pipeline.py
@@ -1080,6 +1080,33 @@ class Pipeline:
         args, kwargs
             model-specific parameters (like paths, formats, etc)
         """
+        if mode == 'static':
+            self.models.load_model(mode, name, model_class, *args, **kwargs)
+            return self
+        return self._add_action(LOAD_MODEL_ID, *args,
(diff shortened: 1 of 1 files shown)


roman-kh:
It was meant to load at each iteration. There are `pipeline.before.load()` for initial loading and `pipeline.load()` for recurring loading as you might want to reload model trained in a separate pipeline or even process.

Perhaps we could introduce a parameter to manage the action (once or at each iteration) or a separate action (e.g. `load_once`) and describe the difference clearly in the docstring.

The very same concerns `save` action as well.

batchflow/pipeline.py
```diff
@@ -1087,6 +1087,29 @@ def load_model(self, mode, name=None, model_class=None, *args, **kwargs):
                                 _args=dict(mode=mode, model_class=model_class, model_name=name),
                                 **kwargs)
 
+    def load_model_once(self, mode, name=None, model_class=None, *args, **kwargs):
+        """ Load a model once at the first iteration
```
roman-kh, line 1091:
```suggestion
        """ Load a model once before the first iteration
```

batchflow/pipeline.py
```diff
@@ -1133,6 +1156,23 @@ def save_model(self, name, *args, **kwargs):
         """
         return self._add_action(SAVE_MODEL_ID, *args, _args=dict(model_name=name), **kwargs)
 
+    def save_model_once(self, name, *args, **kwargs):
+        """ Save a model at the last iteration
```
roman-kh, line 1160:
```suggestion
        """ Save a model after the last iteration
```

roman-kh (changes requested):
Update docstrings for `load` ad `save` as well

batchflow/pipeline.py
```diff
@@ -1133,6 +1156,23 @@ def save_model(self, name, *args, **kwargs):
...
 
+    def save_model_once(self, name, *args, **kwargs):
+        """ Save a model after the last iteration
+
+        Parameters
+        ----------
+        name : str
+            a model name
+
+        batch : Batch
```
roman-kh, line 1167: There is no `batch`

batchflow/pipeline.py
```diff
@@ -1087,6 +1087,29 @@ def load_model(self, mode, name=None, model_class=None, *args, **kwargs):
...
+        mode : str
+            'static' or 'dynamic'
+
+        name : str
+            (optional) a model name
+
+        model_class : class or named expression
+            (optional) a model class to instantiate a loaded model instance.
+
+        batch : Batch
```
roman-kh, line 1104: There is no `batch`

roman-kh (changes requested):

PR #541, opened 2020-11-26
Fix P named expression

roman-kh:


batchflow/named_expr.py
```diff
@@ -515,37 +518,40 @@ class F(NamedExpression):
...
-    def __init__(self, name, mode='w', _pass=True):
-        super().__init__(name, mode)
-        self._pass = _pass
-        self._call = True
-
-    def get(self, **kwargs):
-        """ Return a value from a callable """
+    def get(self, _pass=True, _call=True, **kwargs):
+        """ Return a value from a callable
+        _pass : bool
```
SergeyTsimfer, line 523: If I am not mistaken, you can use `F(my_func)(B())` to do the same, but explicitly

roman-kh: `L(my_func)(B())` to be precise, but yes. `F` was introduced before `B` was created, but later remained for brevity and backward compatibility (which sounds rather funny when I say it out loud).
That is why I'm thinking of refactoring external function calls in pipeline with or without named expressions a little bit later.

PR #545, opened 2021-01-12
Pass microbatches on device individually

SergeyTsimfer:
This PR proposes to split `TorchModel._fill_input` method, which formerly did the following:

- parsed positional and named arguments into model `inputs` and `targets`
- transfered them to `model.device`

into two methods: `parse_inputs` and `transfer_to_device`. That allows for finer control over when data is copied onto GPU, which is necessary to pass individual microbatches at a time (not the whole training batch).

It also allows for more options for passed data: now model can be trained with `np.ndarray`'s, `torch.Tensor`'s and `cupy.ndarray`'s.

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -881,7 +881,9 @@ class TorchModel(BaseModel, VisualizationMixin):
                     self.classes = self.target_shape[1]
 
             self.build_config()
-            self._build([item[:2] for item in split_inputs[0]])
+            build_inputs = [item[:2] for item in split_inputs[0]]
+            build_inputs = self.transfer_to_device(build_inputs)
+            self._build(build_inputs)
             self.model_lock.release()
 
         self.model.train()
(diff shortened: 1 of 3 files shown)


batchflow/models/torch/base.py
```diff
@@ -843,28 +860,30 @@ def train(self, *args, feed_dict=None, fetches=None, use_lock=True, profile=Fals
...
 
-            self.target_shape = get_shape(splitted_targets[0])
+            self.target_shape = get_shape(split_targets[0])
             if self.classes is None:
                 if len(self.target_shape) > 1: # segmentation
                     self.classes = self.target_shape[1]
 
             self.build_config()
-            self._build([item[:2] for item in splitted_inputs[0]])
+            build_inputs = [item[:2] for item in split_inputs[0]]
```
a-arefina, line 884: please add some comment on why we use `item[:2]`

PR #546, opened 2021-01-12
SAM

SergeyTsimfer:
This PR implements the [Sharpness-Aware Minimization for Efficiently Improving Generalization](https://arxiv.org/pdf/2010.01412v2.pdf) paper inside the `TorchModel.train` method. 

Code is heavily inspired by [this](https://github.com/davda54/sam/blob/main/sam.py) and [this](https://github.com/moskomule/sam.pytorch/blob/main/sam.py) repositories.

Currently, this PR missing:

- [x] documentation
- [x] proper naming for turning on the SAM logic and its parameters (I really dislike the `sam` parameter)
- [x] (after updating to `torch >= 1.7.0`): change list comprehensions to in-place operations
- [ ] tests and benchmarks on big datasets with multiple models

Despite the great success of the article (this method sets a number of SOTAs on multiple CV-related tasks), it would clutter the `train` method a lot. Therefore, we should decide, whether we need this functionality in the base `TorchModel`, or we could move it to some other place and use it on demand.

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -204,7 +204,6 @@ class TorchModel(BaseModel, VisualizationMixin):
         If True, then each gradient is scaled according to its own L2 norm.
         If False, then one common gradient norm is computed and used as a scaler for all gradients.
 
-
     order : sequence
         Defines sequence of network blocks in the architecture. Default is initial_block -> body -> head.
         Each element of the sequence must be either a string, a tuple or a dict.
@@ -353,7 +352,7 @@ class TorchModel(BaseModel, VisualizationMixin):
         # Store some of the config values
         self.microbatch = self.full_config.get('microbatch', None)
         self.sync_frequency = self.full_config.get('sync_frequency', 1)
-        self.sam_rho = self.full_config.get('sam_rho', 0.00)
+        self.sam_rho = self.full_config.get('sam_rho', 0.0)
         self.sam_individual_norm = self.full_config.get('sam_individual_norm', False)
         self.profile = self.full_config.get('profile', False)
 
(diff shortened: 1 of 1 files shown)


batchflow/models/torch/base.py
```diff
@@ -341,6 +343,8 @@ def build(self):
         # Store some of the config values
         self.microbatch = self.full_config.get('microbatch', None)
         self.sync_frequency = self.full_config.get('sync_frequency', 1)
+        self.sam_parameter_placeholder = self.full_config.get('sam_parameter_placeholder', False)
```
roman-kh, line 346: Why not just `sam` for brevity?

roman-kh:
It does not seem cluttered for me.

batchflow/models/torch/base.py
```diff
@@ -942,16 +954,54 @@ def train(self, *args, feed_dict=None, fetches=None, use_lock=True, profile=Fals
...
+            #pylint: disable=protected-access
+            # Use obtained grad to move to the local maxima
+            grads = []
+            params_with_grads = []
+            for p in self.model.parameters():
+                if p.grad is not None:
+                    grads.append(p.grad.clone().detach())
+                    params_with_grads.append(p)
+                    p.grad = None
+            grad_norm = torch.stack([g.detach().norm(2).to(self.device) for g in grads]).norm(2)
```
a-arefina, line 985: do you really need to compute norm inside the list comprehension?

batchflow/models/torch/base.py
```diff
@@ -942,16 +954,54 @@ def train(self, *args, feed_dict=None, fetches=None, use_lock=True, profile=Fals
...
+            epsilons = [eps * sam_rho / grad_norm for eps in grads]
```
a-arefina, line 986: What about making `grads` a tensor and scaling it 'simultaneously', instead of iterating over a list?

batchflow/models/torch/base.py
a-arefina: what about just having sam_rho == 0 for no SAM?

roman-kh (approved):

PR #549, opened 2021-01-22
force_parallel param in P NamedExpr

a-arefina:
This PR introduces `force_parallel` argument to `P` Named Expression

this parameter distinguishes 2 cases: 
* with  `force_parallel=False`, which is the default, `P` should be initialized with an iterable with the length equal to batch size. `P` uses each element of this iterable for different batch items
* with `force_parallel=True`, `P` recalculates its argument for each batch item, that is, former `P(R(...))` expressions should be changed to `P(R(...), force_parallel=True)`

This change was motivated by inability to use complex expressions as `P` arguments, ex. `P(R(...)+R(...))`

### DIFF
--- a/batchflow/named_expr.py
+++ b/batchflow/named_expr.py
@@ -722,7 +722,7 @@ class P(W):
     Each image in the batch will be rotated at its own angle::
 
         pipeline
-            .rotate(angle=P(R('normal', 0, 1)))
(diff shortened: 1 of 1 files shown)


roman-kh:
It'll be fixed in `fix-p` branch

roman-kh:
Fixed in #554

PR #554, opened 2021-02-02
Better parallelism with P and PP

roman-kh:
Rename `L` to `F`
Add `PP` for paralleling single-value expressions 
Make proper `repr` for NE
Support `defaultdict` in NE
Add pipeline locks


docs/intro/named_expr.rst
```diff
@@ -14,10 +14,11 @@ There are several types of named expressions:
 * V('name') - a pipeline variable name
 * C('name') - a pipeline config option
 * D('name') - a dataset attribute
-* F(...) - a callable which takes a batch (could be a batch class method or an arbitrary function)
+* F(...) - a callable
 * R(...) - a random value
 * W(...) - a wrapper for a named expression
 * P(...) - a wrapper for parallel actions
+* PP(...) - a wrapper for single-value expressions for parallel actions
```
a-arefina, line 21:
```suggestion
* P(...) - a wrapper for parallel actions that distributes its argument's items among batch items
* PP(...) - a wrapper for single-value expressions for parallel actions that recalculates its argument's expression for each batch item
```

batchflow/tests/named_expr_test.py
```diff
@@ -94,16 +99,16 @@ def test_apply_parallel_p(named_expr, src):
     else:
         assert (b.images == b.masks).all()
 
-
+@pytest.mark.parametrize('p_type', P_OPTIONS)
 @pytest.mark.parametrize('named_expr', P_NAMED_EXPRS)
-def test_inbatch_parallel_p(named_expr):
+def test_inbatch_parallel_p(p_type, named_expr):
```
a-arefina, line 104: this test checks only that the pipeline doesn't fail

It does not check that NE's are calculated properly

I think these tests need thorough rework

batchflow/tests/named_expr_test.py
```diff
@@ -94,16 +99,16 @@ def test_apply_parallel_p(named_expr, src):
     else:
         assert (b.images == b.masks).all()
```
a-arefina, line 100: can't understand how this test passes, because b.images are supposed to be randomly changed in `ap_test`...

batchflow/named_expr.py
```diff
@@ -769,7 +802,7 @@ def get(self, *args, parallel=False, **kwargs):   # pylint:disable=arguments-dif
             return self.name
 
         # pre-calculate values to pass them into decorator which takes them one by one
-        if isinstance(name, R):
+        if isinstance(name, (R, AlgebraicNamedExpression)):
```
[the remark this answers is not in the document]
roman-kh: NE's ``get`` method should accept ``kwargs`` and process only those it needs. So it should work and that is what the test is for.

batchflow/tests/named_expr_test.py
roman-kh: The very same transformation is applied to both `images` and `masks`

roman-kh: And the case where only `images` are transformed is checked in other branch

PR #555, opened 2021-02-02
fix_ci

nikita-klsh:
Batchflow is now being successfully installed from the fork via pip during installation workflow.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -64,7 +64,7 @@ jobs:
       run: |
         pip install --user -U pip 
         pip install wheel
-        pip install -e git+https://github.com/${{ github.repository }}.git@${{ github.head_ref }}#egg=batchflow
+        pip install -e git+https://github.com/${{ github.event.repository.full_name  }}.git@${{ github.head_ref }}#egg=batchflow
     
     - name: Run 'import batchflow'
       run: python -c 'import batchflow' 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-install.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       run: |
         pip install --user -U pip 
         pip install wheel
-        pip install -e git+https://github.com/${{ github.repository }}.git@${{ github.head_ref }}#egg=batchflow
+        pip install -e git+https://github.com/${{ github.event.pull_request.head.repo.full_name }}.git@${{ github.head_ref }}#egg=batchflow
```
roman-kh, line 67: We don't seem to need `-e` option here. Perhaps `--user` is more suitable.

.github/workflows/test-install.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         pip install --user -U pip 
         pip install wheel pipenv
         pipenv --python ${{ matrix.python-version }}
-        pipenv install -e git+https://github.com/${{ github.repository }}@${{ github.head_ref }}#egg=batchflow
+        pipenv install -e git+https://github.com/${{ github.event.pull_request.head.repo.full_name }}.git@${{ github.head_ref }}#egg=batchflow
```
roman-kh, line 103: We don't seem to need `-e` option here.

PR #559, opened 2021-02-06
Torch improvements

SergeyTsimfer:
This PR proposes to:

- use the `padding` parameter of `Conv#D` layers instead of manual `F.pad` in the majority of the cases: if the padding is either one `int`, a tuple of `ints` (i.e. `(2, 3)`), a tuple of tuples of `ints` with the same values (i.e. `((2, 2), (3, 3))`)
- `lock` the entire `train` method
- add `AMP` - automated mixed precision in `train`, which provides 15-30% speed-up depending on the model architecture
- make methods for creating `loss`, `optimizer`, `decay` public, change their signatures and make them modify corresponding attributes in-place
- remove the possibility to pass a list of losses into the model configuration
- add `set_model` method to set external `Torch` module as the `model`, while also updating `optimizer` and `decay`
- add some of the missing docstrings

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -185,6 +185,9 @@ class TorchModel(BaseModel, VisualizationMixin):
         Whether to collect stats of model training timings.
         If True, then stats can be accessed via `profile_info` attribute or :meth:`.show_profile_info` method.
 
+    amp : bool
+        Whether to use automated mixed precision during model training.
+
     sync_frequency : int
         How often to apply accumulated gradients to the weights. Default value is to apply them after each batch.
 
@@ -311,12 +314,17 @@ class TorchModel(BaseModel, VisualizationMixin):
         self.decay = None
         self.decay_step = None
 
+        self.scaler = None
+        self.amp = True
+
         self.callbacks = []
 
         # Memory amortization: accumulate gradients to update weights later
         self.sync_frequency = 1
         self.sync_counter = 0
         self.microbatch = None
+
+        # Sharpness-aware minimization
         self.sam_rho = 0.0
         self.sam_individual_norm = True
 
@@ -352,6 +360,8 @@ class TorchModel(BaseModel, VisualizationMixin):
         # Store some of the config values
         self.microbatch = self.full_config.get('microbatch', None)
         self.sync_frequency = self.full_config.get('sync_frequency', 1)
+        self.amp = self.full_config.get('amp', True)
+
         self.sam_rho = self.full_config.get('sam_rho', 0.0)
         self.sam_individual_norm = self.full_config.get('sam_individual_norm', False)
         self.profile = self.full_config.get('profile', False)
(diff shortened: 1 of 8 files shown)


batchflow/models/torch/base.py
```diff
@@ -539,7 +560,7 @@ def _get_placeholder_shapes(self):
 
 
     # Chain multiple building blocks to create model
-    def _build(self, inputs=None):
+    def _build_model(self, inputs=None):
```
roman-kh, line 563: model._build_model() is rather verbose

batchflow/models/torch/base.py
```diff
@@ -648,50 +665,75 @@ def _make_optimizer(self, config):
...
+            self.decay.append(decay_)
+            self.decay_step.append(step_params)
 
-            decays.append(decay)
-            list_kwargs.append(decay_args)
-            list_steps.append(step_meta)
-        return decays, list_kwargs, list_steps
 
+    # Use external model
+    def set_model(self, model):
```
roman-kh, line 724: model.set_model() is rather verbose and not obvious. Perhaps `use` or `with` or `create_with`

batchflow/models/torch/base.py
```diff
@@ -982,7 +1037,8 @@ def _train(self, *args, fetches=None, sync_frequency=True, sam_rho=0.0, sam_indi
         inputs = inputs[0] if isinstance(inputs, (tuple, list)) and len(inputs) == 1 else inputs
 
         # Apply model, compute loss and gradients
-        predictions = self.model(inputs)
+        with torch.cuda.amp.autocast(enabled=self.amp):
```
roman-kh, line 1040: what if there is no cuda installed?

roman-kh (changes requested):

batchflow/models/torch/base.py
SergeyTsimfer: On `cpu` version of `PyTorch`, a warning of the `cuda is not installed so not using its features` sort is raised

batchflow/models/torch/base.py
SergeyTsimfer: The naming of `set_model` and `build_model` was used as a preparation for splitting the `TorchModel` class into two parts: one that creates `nn.Module` and one that trains it. I agree that it was too early to rename the `_build` method; can we keep the `set_model` name as it is most descriptive of what the method does?

SergeyTsimfer:
@roman-kh do we want to use `AMP` for predictions?

batchflow/models/torch/base.py
roman-kh: As far as I'm concerned, `set_model` is less descriptive.

batchflow/models/torch/base.py
```diff
@@ -1172,30 +1172,33 @@ def predict(self, *args, targets=None, feed_dict=None, fetches=None, use_lock=Tr
         """
         inputs, targets = self._make_prediction_inputs(*args, targets=targets, feed_dict=feed_dict, **kwargs)
 
-        if use_lock:
-            self.model_lock.acquire()
+        # Acquire lock, release anyway
+        try:
```
roman-kh, line 1176: `try .. except` block might be extremely slow, which is not we want for each iteration.

batchflow/models/torch/base.py
```diff
@@ -1172,30 +1172,33 @@ def predict(self, *args, targets=None, feed_dict=None, fetches=None, use_lock=Tr
...
+                predictions = self.model(inputs)
+                output_container['predictions'] = predictions
 
-            if targets is not None:
-                targets = self.transfer_to_device(targets)
-                loss = self.loss(predictions, targets)
-                output_container['loss'] = loss
+                if targets is not None:
+                    targets = self.transfer_to_device(targets)
+                    loss = self.loss(predictions, targets)
```
roman-kh, line 1190: combine two lines

PR #562, opened 2021-02-10
Process PASCAL labels

AlexeyKozhevin:


batchflow/opensets/pascal.py
```diff
@@ -108,13 +108,19 @@ def download(self, path):
                                for name in  [*train_ids, *test_ids]], dtype=object)
             masks = np.array([self._extract_image(archive, self._mask_path(name)) \
                               for name in [*train_ids, *test_ids]], dtype=object)
+            masks = self._process_mask(masks)
```
roman-kh, line 111: Let's add flag whether to process or not which is True by default

roman-kh (changes requested):

roman-kh:
Why?

PR #573, opened 2021-03-23
Fixes for torch model and notifier

SergeyTsimfer:
The main contribution of this PR is the conversion of predicted arrays back to `float32` precision, even with `amp` turned on

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -186,6 +186,7 @@ class TorchModel(BaseModel, VisualizationMixin):
 
     amp : bool
         Whether to use automated mixed precision during model training and inference. Default is True.
+        The output type of predictions remains float32.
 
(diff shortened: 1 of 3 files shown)


batchflow/models/torch/base.py
```diff
@@ -1183,6 +1183,12 @@ def predict(self, *args, targets=None, feed_dict=None, fetches=None, use_lock=Tr
                 output_container = {}
                 inputs = self.transfer_to_device(inputs)
                 predictions = self.model(inputs)
+
+                if self.amp:
+                    if isinstance(predictions, (tuple, list)):
+                        predictions = [p.float() for p in predictions]
```
roman-kh, line 1189: This might unexpectedly change `predictions` type from `tuple` to `list`.

a-arefina (approved):

PR #574, opened 2021-03-25
Research improvements

AlexeyKozhevin:
Research 2.0

- Changes in architecture
- Execution without dumping of results
- Non-parallel execution
- Each executable unit is a callable or generator
- Change the storage format of the results
- Save current state (commit, diff, status) of the git repository
- Update docstrings

TODO:

- [x] update tutorials
- [x] update converter from old research
- [x] update documentation
- [x] extend tests
- [x] random state
- [x] add namespaces
- [x] add debug mode to use %pdb


batchflow/research/research.py
```diff
@@ -203,6 +209,7 @@ def get_metrics(self, pipeline, metrics_var, metrics_name, *args,
             include execution information to log file or not
         """
         name = pipeline + '_' + metrics_var
+        name += str(len([item for item in self.executables if item.startswith(name)]))
```
a-arefina, line 212: This primitive is used multiple times. I'd make a small function smth like
```
def count_startswith(seq, name): return sum(1 for item in seq if item.startswith(name))
```
or a method with `self.executables` instead of `seq`

batchflow/research/research.py
```diff
@@ -404,6 +412,24 @@ def run(self, n_iters=None, workers=1, branches=1, name=None,
 
         return self
 
+    def attach_git_meta(self, **kwargs):
+        """ Save the information about the current state of project repository. """
```
a-arefina, line 416: please add kwargs description in docstring

batchflow/research/results.py
```diff
@@ -78,16 +50,219 @@ class Results:
...
+        self.configs = self._load_configs()
+        self.names = list(self.description['executables'].keys())
+        self.variables = [var for unit in self.description['executables'].values() for var in unit['variables']]
 
-    def _get_list(self, value):
-        if not isinstance(value, list):
-            value = [value]
-        return value
+    @property
+    def df(self):
```
a-arefina, line 61: not quite sure that we'll need this property now. 
It silently starts a very heavy loading and think it is better to make a user explicitly state what they want to load via `load_df` arguments

batchflow/research/results.py
```diff
@@ -78,16 +50,219 @@ class Results:
...
+
+    def _load_configs(self, experiment_id=None):
+        configs = {}
+        for filename in glob.glob(os.path.join(self.path, 'configs', experiment_id or '*')):
+            _experiment_id = os.path.split(filename)[-1]
+            with open(filename, 'rb') as f:
+                configs[_experiment_id] = dill.load(f)
+        return configs
+
+    def _to_list(self, value):
```
a-arefina, line 264: make it static, or move to some utility module

batchflow/research/results.py
```diff
@@ -127,16 +302,18 @@ def _concat(self, results, variables):
...
+        """ Pad arrays in the dict with nans to the same length. """
+        max_len = max([len(value) if isinstance(value, (list, tuple, pd.Series)) else 1 for value in chunk.values()])
         for value in chunk.values():
             if len(value) < max_len:
-                value.extend([pd.np.nan] * (max_len - len(value)))
+                value.extend([np.nan] * (max_len - len(value)))
+        return max_len
 
-    def _filter_configs(self, config=None, alias=None, repetition=None):
+    def _filter_configs(self, _configs, config=None, alias=None, repetition=None):
```
a-arefina, line 312: what's the difference with `filter_configs` above?

batchflow/models/torch/base.py
```diff
@@ -999,10 +999,16 @@ def train(self, *args, feed_dict=None, fetches=None, use_lock=True, profile=Fals
 
             # Parse outputs to a desired structure
             if fetches:
```
a-arefina, line 1001: what we really need to make any changes in this code is a list of possible output contents  and a test suite that covers it.

a-arefina (commented):
See some minor comments on this PR.
I am really happy with a method that converts old results format to a new one!
And I also have a tiny feature request: please add a __len__ method to `Option` class. 

as for `fetches` fix, we need unit tests to cope with it, I'm afraid

batchflow/research/results.py
AlexeyKozhevin: At least we need it for backward compatibility.

batchflow/models/torch/base.py
AlexeyKozhevin: Agree. But this requires the code to work, so these changes were made. Before this, saving multiple fetches to different destinations did not work

batchflow/research/job.py
```diff
@@ -18,17 +20,23 @@ def __init__(self, executable_units, n_iters, configs, branches, research_path):
...
-        self.ids = [str(random.getrandbits(32)) for _ in self.configs]
+        self.ids = [self._make_random_name(config[0]) for config in self.configs]
 
         self.exceptions = []
         self.stopped = []
         self.last_update_time = None
 
+    def _make_random_name(self, config):
+        name = hashlib.md5(config.alias(as_string=True).encode('utf-8')).hexdigest()
+        name += str(random.getrandbits(16))
```
SergeyTsimfer, line 37: Why do you need this randomization? Hash collision is improbable

batchflow/research/job.py
```diff
@@ -38,29 +46,53 @@ def init(self, worker_config, device_configs, last_update_time):
...
             self.exceptions.append(None)
         self.clear_stopped_list()
 
+    def create_folders(self):
+        """ Create folder if it doesn't exist """
+        self.experiment_paths = [os.path.join('results', experiment_id) for experiment_id in self.ids]
+        for path in self.experiment_paths:
+            try:
+                os.makedirs(os.path.join(self.research_path, path))
+            except FileExistsError:
```
SergeyTsimfer, line 93: There is `exists_ok` parameter for `os.makedirs`

batchflow/research/research.py
```diff
@@ -19,14 +22,32 @@
...
-from .utils import get_metrics
-from .executable import Executable
+from .utils import get_metrics, count_startswith
+from .executable import Executable, Namespace
 from .named_expr import RP
 
 class Research:
     """ Class Research for multiple parallel experiments with pipelines. """
-    def __init__(self):
+    def __init__(self, domain=None, n_configs=None, n_reps=1, repeat_each=100):
```
SergeyTsimfer, line 31: As this signature is actually visible and tabbable in notebooks, let's add more documentation here (at least for other parameters)

batchflow/research/research.py
```diff
@@ -172,6 +195,26 @@ def add_callable(self, function, *args, returns=None, name=None, execute=1, dump
...
+        Parameters
+        ----------
+        namespace : class
+            class which represents namespace.
+        name : str
+            name of the namespace to use in research.
+        """
+        if name in self.namespaces:
+            raise ValueError('Namespace with name {} already exists'.format(name))
+        if name is None:
```
SergeyTsimfer, line 210: Is that branch really needed? I believe that it is better for things as `init_something` to be as explicit as possible, especially regarding the naming, which one will later need to add actions to research

batchflow/research/research.py
```diff
@@ -203,32 +246,13 @@ def get_metrics(self, pipeline, metrics_var, metrics_name, *args,
...
         """
         name = pipeline + '_' + metrics_var
+        name += str(len([item for item in self.executables if item.startswith(name)]))
         returns = returns or metrics_name
         self.add_callable(get_metrics, *args, name=name, execute=execute, dump=dump, returns=returns,
                           on_root=False, logging=logging, pipeline=RP(pipeline),
                           metrics_var=metrics_var, metrics_name=metrics_name, agg=agg, **kwargs)
         return self
 
-    def init_domain(self, domain=None, n_configs=None, n_reps=1, repeat_each=100):
```
SergeyTsimfer, line 212: Maybe, you can leave this function, and call it in `__init__` if the domain was passed at initialization?

batchflow/research/utils.py
```diff
@@ -1,6 +1,15 @@
 """ Auxilary functions """
+import os
+import glob
+import shutil
 
 
+def to_list(value):
```
SergeyTsimfer, line 7: @alexandrkuvaev

batchflow/research/results.py
```diff
@@ -78,16 +52,232 @@ class Results:
...
+        -------
+        pandas.DataFrame or dict
+            dataframe columns: iteration, name (of pipeline/function), repetition, update
+            (is used if domain was dynamically updated), and column(-s) for config.
+            Also it will have column for each variable of pipeline
+            and output of the function that was saved as a result of the research.
+        """
+        return self._load_df(*args, **kwargs)
+
+    def load_artifactes(self, *args, **kwargs):
```
SergeyTsimfer, line 108: Why is it artifact**e**s?

batchflow/research/results.py
```diff
@@ -5,44 +5,18 @@
 import glob
 import json
 import dill
+import numpy as np
 import pandas as pd
 
+from .utils import to_list
+
 class Results:
```
SergeyTsimfer, line 13: In my researches, if some class (for example, neural network block written as `nn.Module`) defined in the notebook itself, it raises errors on results loading. Is it possible to deal with it?

AlexeyKozhevin:
Research 2.0

batchflow/research/domain.py
```diff
@@ -422,51 +468,54 @@ def _iterator():
...
 
-    @property
-    def size(self):
-        """ The number of configs that will be produces from domain. """
-        if self.n_iters is not None:
-            return self.n_reps * self.n_iters
-        return None
-
-    @property
     def iterator(self):
```
SergeyTsimfer, line 488: Definitely looks like `iterator` should be a property

batchflow/research/distributor.py
```diff
@@ -1,181 +1,232 @@
...
+        self.research.logger.info(f'Get {n_tasks} tasks with {n_configs} configs, remains {self.configs_remains}')
+
+        return n_configs
+
+    def stop_workers(self, n_workers):
+        """ Stop all workers by putting `None` task into queue. """
+        for _ in range(n_workers):
+            self.put(None)
+
+    def join(self):
```
SergeyTsimfer, line 71: `__getattr__`, maybe?

SergeyTsimfer: Better yet, just inherit from `mp.JoinableQueue`

SergeyTsimfer (commented):
Most of the classes need docstring written from the scratches, with little to no connection and references to other classes. Current documentation already expects the reader to know a lot of things about `Research`
Also, what do you think about adding `experiment_config` variable to the scope of all callables? So that I could use it without explicitly mentioning it in the `add_callable` signature

batchflow/research/domain.py
```diff
@@ -126,24 +131,29 @@ def iterator(self):
 
 
 class ConfigAlias:
-    """ Class for config and alias which is represenation of config where all keys are str.
+    """ Class for config and its aliased version where all keys and values are `str`.
```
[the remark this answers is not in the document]
SergeyTsimfer: That is supposed to be a comment for `Option`. Also, do we really need a separate `Option` class in the first place?

batchflow/research/experiment.py
```diff
@@ -0,0 +1,785 @@
...
+        self.outputs = dict()
+        self.results = OrderedDict()
+
+        self.has_dump = False # does unit has any dump actions or not
+        self.name = None # name of the executor/research
+        self.dump_results = None
+        self.loglevel = None
+        self.monitor = None
+
+        self.id = None #pylint:disable=invalid-name
```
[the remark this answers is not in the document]
AlexeyKozhevin: To distinguish two states of Experiment: experiment as a scheme (without config) and experiment as executable item (with config). `id` can exist only for the second.

batchflow/research/experiment.py
```diff
@@ -0,0 +1,785 @@
...
+        self.add_callable(name, _dump_results,
+                          iterations_to_execute=iterations_to_execute,
+                          variable=variable, experiment=E())
+        return self
+
+    def add_postfix(self, name):
+        """ Add postfix for conincided unit name. """
+        return name + '_' + str(sum([item.startswith(name) for item in self.actions]))
+
+    def only_callables(self):
```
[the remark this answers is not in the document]
AlexeyKozhevin: Not really: `is_alive` can't change value `False` to `True` but `only_callables` can. For example, you can remove some action from `Experiment` (but don't do that )

batchflow/research/research.py
```diff
@@ -442,173 +176,350 @@ def _get_devices(self, devices):
...
+                elif status == 'START_EXP':
+                    self.current_iterations[signal['id']] = 0
+                elif status == 'EXECUTE_IT':
+                    self.current_iterations[signal['id']] = signal['it']
+                elif status == 'FINISH_EXP':
+                    self.current_iterations.pop(signal['id'])
+                    self.finished_iterations += signal['it'] + 1
+                    self.finished_experiments += 1
+
+                if status in ['START_EXP', 'EXECUTE_IT', 'FINISH_EXP']:
```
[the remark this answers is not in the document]
AlexeyKozhevin: Because for `update` you have to compute deltas and it doesn't add clarity

batchflow/research/distributor.py
AlexeyKozhevin:
```
import multiprocess as mp
type(mp.JoinableQueue)
>>> method
```

PR #580, opened 2021-04-29
Allow DataParallel mode for loaded models

dpodvyaznikov:
* `TorchModel` now has `_to_device` method that switches between `DataParallel` and single-device modes based on a number of provided devices.
* torch models now can be loaded in `DataParallel` mode if `device` key of load config is a list of devices.


SergeyTsimfer (approved):

batchflow/models/torch/base.py
```diff
@@ -1402,11 +1401,16 @@ def load(self, path, *args, eval=False, **kwargs):
         else:
             checkpoint = torch.load(path, **kwargs)
 
+        # Save load config values to update `full_config` later
+        load_config = self.config
+        # Remove unnecessary key from load config
+        _ = load_config.pop('build')
```
roman-kh, line 1407: `load_config` is a reference to `self.config`

batchflow/models/torch/base.py
```diff
@@ -559,6 +559,11 @@ def _get_placeholder_shapes(self):
                 if self.classes is None:
                     self.classes = shapes[0][0]
 
+    def _to_device(self):
+        if len(self.devices) > 1:
+            self.model = nn.DataParallel(self.model, self.devices)
```
roman-kh, line 564: Do we always need data parallel with multiple devices?

batchflow/models/torch/base.py
```diff
@@ -559,6 +559,11 @@ def _get_placeholder_shapes(self):
                 if self.classes is None:
                     self.classes = shapes[0][0]
 
+    def _to_device(self):
```
roman-kh, line 562: Why is it private?

dpodvyaznikov: There is no obvious reason to make it public in this form.

It is possible to make it public and change its signature to `to_device(self, device=None)`, but in this case we also need to change `self.device` and process all the model-related object, e.g. move loss to another device if it is an instance of `nn.Module`, reconstruct optimizers. 

Do you see cases when you might want to call this method independently and move a constructed model to another device on the fly?

batchflow/models/torch/base.py
dpodvyaznikov: For now, I believe we do.
The only exception I see is when you want to run model parallel mode, e.g. you have have a model that doesn't fit into one device and you want to put it on several devices.

batchflow/models/torch/base.py
dpodvyaznikov: `self.config` is overwritten in line 1410, when corresponding attr is set, but I want to update `full_config` with `self.config`'s values to update `'device'` and save loading path.

I'll remove `pop` from here since it's actually unnecessary and expand the comment

batchflow/models/torch/base.py
roman-kh: Reflect all these comments in the docstring, please

PR #587, opened 2021-06-04
Add light profiling

roman-kh:


batchflow/pipeline.py
```diff
@@ -1474,7 +1485,12 @@ def reset(self, *args, profile=False, seed=None):
 
         self.random_seed = seed
 
-        self._profiler = Profile() if profile else None
+        if profile == 2 or 'detailed'.startswith(profile):
```
AlexeyKozhevin, line 1488: If `profile` is bool that check will fail.

PR #588, opened 2021-06-04
BA component

anton-br:
The pull request adding a named expression - __BA__ that uses when batch component contains an array of objects. The component redirect calls to every object in the batch.

### DIFF
--- a/batchflow/named_expr.py
+++ b/batchflow/named_expr.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 import numpy as np
 
 from .config import Config
+from .utils import to_list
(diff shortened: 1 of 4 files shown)


batchflow/named_expr.py
```diff
@@ -434,6 +437,48 @@ def assign(self, value, **kwargs):
         if name is not None:
             setattr(batch, name, value)
 
+class BA(B):
+    """ Array batch component. Uses when the batch component is an array of objects.
```
roman-kh, line 441:
```suggestion
    """ Array component to use when the batch component is an array of objects.
```

PR #592, opened 2021-06-21
Updates in losses and `build_config` for `TorchModel`

HollowPrincess:


SergeyTsimfer (commented):
Update the branch to current master

batchflow/models/torch/base.py
```diff
@@ -477,12 +477,12 @@ def build_config(self, names=None):
             config['common/data_format'] = config.get('common/data_format') or data_format
 
         config['head/target_shape'] = self.target_shape
-        config['head/classes'] = config.get('head/classes', self.classes)
+        config['head/classes'] = config.get('head/classes') or self.classes
```
roman-kh, line 480:
```suggestion
        config['head/classes'] = config.get('head/classes', self.classes)
```
is better (in particular for `classes=False`)

HollowPrincess: In this case, we found bad behavior in model initialization in the extension task. This code runs twice for the task: the first time it doesn't know any information about ```classes``` and key ```config['head/classes'] ``` sets as ```None```.  At the second run, the code doesn't rewrite the value for the key ```head/classes``` here, because it was set before with ```None``` value.

PR #600, opened 2021-07-08
10 full project tutorial

HollowPrincess:


batchflow/models/torch/blocks.py
```diff
@@ -112,7 +112,9 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
         kernel_size = [kernel_size] * num_convs if isinstance(kernel_size, int) else kernel_size
         strides = [strides] * num_convs if isinstance(strides, int) else strides
         groups = [groups] * num_convs
-        branch_stride = np.prod(strides)
+        branch_params = kwargs.get('branch_params', {})
```
roman-kh, line 115: Why not just `branch`?

batchflow/notifier.py
```diff
@@ -387,6 +387,7 @@ def create_description(self, iteration):
 
     def create_message(self, iteration, description):
         """ Combine timestamp, iteration and description into one string message. """
+        iteration = iteration//self.frequency
```
SergeyTsimfer, line 390: Why?

batchflow/models/torch/resnet.py
```diff
@@ -115,7 +115,7 @@ class ResNet50(ResNet34):
     @classmethod
     def default_config(cls):
         config = super().default_config()
-        config['body/encoder/blocks/layout'] = 'cna'
+        config['body/encoder/blocks/layout'] = 'cn'
```
SergeyTsimfer, line 118: Why?

batchflow/models/torch/blocks.py
```diff
@@ -112,7 +112,9 @@ def __init__(self, inputs=None, layout='cnacn', filters='same', kernel_size=3, s
         kernel_size = [kernel_size] * num_convs if isinstance(kernel_size, int) else kernel_size
         strides = [strides] * num_convs if isinstance(strides, int) else strides
         groups = [groups] * num_convs
-        branch_stride = np.prod(strides)
+        branch_params = kwargs.pop('branch', {})
```
SergeyTsimfer, line 115: Add to signature

batchflow/notifier.py
HollowPrincess: A timestamp list saves information by frequency, and when the frequency is more than 1 iteration, we have more iterations than timestamps. To catch the right element in a timestamps list, we need to get the right index.

batchflow/models/torch/resnet.py
HollowPrincess: A bottleneck parameter is set to ```True```, which makes a ResBlock layout equals to ```'cna'``` + config layout + ```'acn'``` (from ResBlock ```__init__``` method).

batchflow/models/metrics/classify.py
```diff
@@ -158,6 +159,35 @@ def __getattr__(self, name):
...
+        classes : sequence, optional
+            Sequence of classes labels.
+        normalize : bool
+            Whether to normalize confusion matrix over target classes.
+        """
+        confusion_matrix = np.array(self.confusion_matrix)
+        if classes is None:
+            classes = np.arange(self.num_classes)
+        if normalize:
+            confusion_matrix = confusion_matrix/np.nansum(confusion_matrix, axis=0)
```
cdtn, line 176:
```suggestion
            confusion_matrix = confusion_matrix / np.nansum(confusion_matrix, axis=0)
```

batchflow/notifier.py
cdtn:
```suggestion
        iteration = iteration // self.frequency
```

batchflow/opensets/imagenette.py
```diff
@@ -59,6 +61,8 @@ def _is_file_rgb(archive, member):
             """ Check whether archive member is a file.
             In case `drop_grayscale` set to `True` it verifies that the member is the RGB mode image as well.
             """
+            if (member.name.find('csv')!=-1) or (member.name.find('.DS_Store')!=-1):
```
cdtn, line 64:
```suggestion
            if (member.name.find('csv') != -1) or (member.name.find('.DS_Store') != -1):
```

batchflow/models/metrics/classify.py
```diff
@@ -158,6 +159,35 @@ def __getattr__(self, name):
...
+
+        fig, ax = plt.subplots(figsize=(10, 10))
```
anton-br, line 178: why do you hardcode figsize?

batchflow/models/torch/base.py
```diff
@@ -316,6 +316,7 @@ def __init__(self, config=None):
         self.devices = []
 
         # Train procedure and ifrastructure
+        self.init_model_weights = None
```
anton-br, line 319: mention that it is a kaiming normal initialization. Rename to `kaiming_normal_init` or etc.

HollowPrincess: Ok, the method is renamed to ```init_kaiming_normal``` (like torch style), the flag renamed to ```kaiming_normal_init```.

HollowPrincess: Also, I added the ```init_zero_bias``` flag to separate these initializations.

batchflow/models/metrics/classify.py
HollowPrincess: I added it to parameters from now, but I am not sure that it is the only parameter we need to add.

batchflow/notifier.py
HollowPrincess: @cdtn, applied

batchflow/models/torch/base.py
```diff
@@ -316,6 +316,8 @@ def __init__(self, config=None):
         self.devices = []
 
         # Train procedure and ifrastructure
+        self.kaiming_normal_init = None
```
SergeyTsimfer, line 319: Can you move `init` to a separate PR? I reckon it would be merged a touch faster, than the tutorial

SergeyTsimfer: What about other way of applying initialization strategies: a key `init` in the config, that can be used to pass a string (e.g. `default`) to choose a pre-defined initialization strategy (a callable on nn.Module), or a callable that will be applied directly?

batchflow/models/utils.py
```diff
@@ -10,6 +10,8 @@ def unpack_args(args, layer_no, layers_max):
                 arg_value = args[arg][layer_no]
             else:
                 arg_value = args[arg]
+        elif isinstance(args[arg], dict): # for args with dict-like structure, e.g. branch in ResBlock
```
SergeyTsimfer, line 13: This comment really does not help to understand the use case. Also, this function is used in other places, and it may be very sensitive to such changes...

batchflow/opensets/imagenette.py
```diff
@@ -29,10 +29,12 @@ class Imagenette(ImagesOpenset):
 
     """
     SOURCE_URL = 'https://s3.amazonaws.com/fast-ai-imageclas/imagenette.tgz'
+    CLASSES = ['tench', 'English springer', 'cassette player', 'chain saw', 'church',
```
SergeyTsimfer, line 32: woof

batchflow/models/torch/base.py
HollowPrincess: It is a good idea. Also, this PR contains some fixes and additions; should they be separated too?

batchflow/models/utils.py
HollowPrincess: Without it, we can't nicely parse something like:
```
config = {
    ...
    'branch/layout' : [...]
    ...
}
```
(A problem is a list in a dict)

Have you got any other **suggestions**? (except hardcode 'branch' key)

PR #603, opened 2021-07-27
Improve trackers

SergeyTsimfer:
* add an option to `disable` `Notifier`
* rename `sp` to `display`, following the `tqdm` API change
* make `GPU*` monitors use better defaults for defining currently used GPU


batchflow/notifier.py
```diff
@@ -91,13 +91,21 @@ class Notifier:
...
     *args, **kwargs
         Positional and keyword arguments that are used to create underlying progress bar.
     """
     def __init__(self, bar=None, *args, update_total=True,
                  total=None, batch_size=None, n_iters=None, n_epochs=None, drop_last=False, length=None,
                  frequency=1, monitors=None, graphs=None, file=None,
-                 window=None, layout='h', figsize=None, savepath=None, **kwargs):
+                 window=None, layout='h', figsize=None, savepath=None, disable=False, **kwargs):
+        # Turn off everything if `disable`
+        if disable:
```
roman-kh, line 104: Why not to add `disable` method?

SergeyTsimfer: It is supposed to be used for external parametrization like:
```python
def my_func(*args, bar=True, **kwargs)
    ...
    notifier = Notifier(..., disable=not bar)
```

batchflow/monitor.py
```diff
@@ -190,6 +198,12 @@ def get_usage(pid=None, **kwargs):
         return process.memory_full_info().uss / (1024 ** 3) # gbytes
 
 
+def get_current_gpus():
+    """ If the `CUDA_VISIBLE_DEVICES` is set, check it and return device numbers. Otherwise, return [0]. """
+    env_variable = os.environ.get('CUDA_VISIBLE_DEVICES', '0')
```
AlexeyKozhevin, line 203: Why '0' instead of all availiable GPUs?

AlexeyKozhevin (approved):

batchflow/monitor.py
SergeyTsimfer: I think it is more often to use only one GPU even if all of them are available in the current notebook (env variable is not set). In cases of multiple GPU, one should manually pass arguments to monitor initialization

PR #606, opened 2021-08-02
Fix default retrieval for `C`

cdtn:
`C.get` fails to retrieve the default value for named expression when pipeline config is empty.

This happens due to a call on line `528`, where keyword argument `default` is passed.
Since the default python `dict` takes no keyword arguments, such call leads to `TypeError`.

### DIFF
--- a/batchflow/named_expr.py
+++ b/batchflow/named_expr.py
@@ -519,7 +519,7 @@ class C(PipelineNamedExpression):
     def get(self, **kwargs):
         """ Return a value of a pipeline config """
         name, pipeline, _ = self._get_params(**kwargs)
-        config = pipeline.config or {}
+        config = pipeline.config or Config()
 
         if name is None:
             return config


batchflow/named_expr.py
```diff
@@ -519,7 +519,7 @@ def __init__(self, name=None, mode='w', **kwargs):
     def get(self, **kwargs):
         """ Return a value of a pipeline config """
         name, pipeline, _ = self._get_params(**kwargs)
-        config = pipeline.config or {}
+        config = pipeline.config or Config()
```
SergeyTsimfer, line 522: That would still fail to work if the pipeline config is a dictionary
It is better to remove the keyword in `get` call

cdtn: Provided config is always converted to `Config` (see `Pipeline.__init__`, line `92`), so dict as a config will ruin nothing.

PR #616, opened 2021-09-03
Research improvements

AlexeyKozhevin:
- speedify profiling: replace `pd.DataFrame` append by list append
- `Research` handles `KeyboardInterrupt`
- fix several callables in research with the same name


batchflow/research/research.py
```diff
@@ -354,28 +354,60 @@ def run(self, name=None, workers=1, branches=1, n_iters=None, devices=None, exec
...
+                self.process = mp.Process(target=_start_distributor)
+                self.process.start()
+                self.logger.info(f"Detach research[pid:{self.process.pid}]")
+            else:
+                _start_distributor()
+                self.terminate()
+        except KeyboardInterrupt as e:
+            self.logger.info("Research has been stopped by KeyboardInterrupt.")
+            self.terminate()
+            raise e
```
SergeyTsimfer, line 373: You don't need to re-raise the `KeyboardInterrupt` error IMO

batchflow/research/research.py
```diff
@@ -354,28 +354,60 @@ def run(self, name=None, workers=1, branches=1, n_iters=None, devices=None, exec
...
         if self.process is not None:
-            self.logger.info(f"Terminate research process[pid:{self.process.pid}]")
-            parent = psutil.Process(self.process.pid)
+            processes_to_kill = [self.process]
+        elif self.distributor is not None:
+            processes_to_kill = self.distributor.worker_processes
+        else:
+            processes_to_kill = []
+
+        for parent in processes_to_kill:
```
SergeyTsimfer, line 390: `process` instead of `parent`, maybe?

batchflow/research/research.py
```diff
@@ -354,28 +354,60 @@ def run(self, name=None, workers=1, branches=1, n_iters=None, devices=None, exec
...
+            parent = psutil.Process(parent.pid)
             for child in parent.children(recursive=True):
+                self.logger.info(f"Terminate process[pid:{child.pid}]")
```
SergeyTsimfer, line 393: Differentiate the logs: `Terminating child process: [parent.pid:], [child.pid]` and `Terminating worker's process: [parent.pid:]`

batchflow/research/experiment.py
```diff
@@ -609,14 +609,14 @@ def dump(self, variable=None, when='last'):
...
         """ Add postfix for conincided unit name. """
-        n_actions = sum(self._is_postifixed(name, item) for item in self.actions)
-        return name if n_actions == 0 else f"{name}_{n_actions}"
+        n_actions = sum(self._is_postifixed(new_name, unit_name) for unit_name in self.actions)
+        return new_name if n_actions == 0 else f"{new_name}_{n_actions}"
 
-    def _is_postifixed(self, base_name, name):
-        postfix = name[len(base_name):]
-        return (name == base_name) or (len(postfix) > 2 and postfix[0] == '_' and postfix[1:].isdigit())
+    def _is_postifixed(self, new_name, unit_item):
```
roman-kh, line 617:
```suggestion
    def _has_postfix(self, new_name, unit_item):
```

batchflow/research/research.py
AlexeyKozhevin: It is difficult to distinguish between the processes:

- if `detach=True` and `parallel=True`: `parent` is detached process and and its children are workers, tasks and possibly their children (we can't destinguish them with `parent.children(recursive=True)`)
- if `detach=True` and `parallel=False`: `parent` is detached process without children
- if `detach=False` and `parallel=True`: `parent` is each of the workers and children are tasks and possibly their children
- if `detach=False` and `parallel=False`: no processes

a-arefina (approved):

batchflow/research/research.py
```diff
@@ -440,6 +516,11 @@ def folder_is_research(cls, name):
             raise FileNotFoundError(f"Folder {name} doesn't exist.")
         return os.path.isfile(os.path.join(name, 'research.dill'))
 
+    @property
+    def is_finished(self):
+        """ The number of experimenys in queue of tasks. """
+        return self.monitor.in_queue + self.monitor.remained_experiments == 0
```
HollowPrincess, line 522: This docstring duplicates the docstring in the `in_queue` property.

batchflow/research/research.py
```diff
@@ -114,33 +118,80 @@ def update_domain(self, function, when, **kwargs):
...
+        commands : dict, optional
+            commands to execute. Each output will saved into '{dst}/{key}.txt'.
         """
+        commands = commands or dict()
         commands = {
             'commit': "git log --name-status HEAD^..HEAD",
             'diff': 'git diff',
             'status': 'git status -uno',
+            **commands,
             **kwargs
```
a-arefina, line 191: What is the difference between `kwargs` and `commands`? Why do you use both? It seems that you really can use *args, **kwargs instead of commands in all this family of functions

PR #623, opened 2021-09-30
refatoring for using subsystems

a-arefina:
ClearML notification V3

code for subsystem
```
class ClearMLNotifierSubsystem(NotifierSubsystem):
    def __init__(self, logger=None):
        self.logger = logger

    def update(self, data_containers, iteration):
        if self.logger:
            for container in data_containers:
                source = container['source']
                if isinstance(source, (str, NamedExpression)):
                    name = container['name']
                    value = container['data'][iteration]
                    self.logger.report_scalar(title=name, series=name, value=value, iteration=iteration)
```

usage

```
cmln = ClearMLNotifierSubsystem(logger=logger)

train_pipeline.run(D('size'), n_iters=N_ITERS,
                   bar = {'bar': 'a', 'monitors': 'loss_history', 'subsystems': [cmln]})
```

### DIFF
--- a/batchflow/notifier.py
+++ b/batchflow/notifier.py
@@ -99,7 +99,7 @@ class Notifier:
     def __init__(self, bar=None, *args, update_total=True,
                  total=None, batch_size=None, n_iters=None, n_epochs=None, drop_last=False, length=None,
                  frequency=1, monitors=None, graphs=None, file=None,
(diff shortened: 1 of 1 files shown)


a-arefina:
we can also have an `add_subsystem`/`register` method in `Notifier`, to allow adding notifications media after Notifier initialization

PR #624, opened 2021-10-01
Improvements

HollowPrincess:


batchflow/models/torch/base.py
```diff
@@ -789,6 +795,104 @@ def head(cls, inputs, **kwargs):
...
 
+    # Model weights initialization
+    def init_weights(self, init_model_weights=None, init_zero_bias=False):
+        """ Initialize model weights with specific distribution and initialize biases to 0.
+
+        About weighs initialization, you can read here: `torch.nn.init <https://pytorch.org/docs/stable/nn.init.html>`_.
+
+        Parameters
+        ----------
+        init_model_weights : dict, True, or None
```
SergeyTsimfer, line 806: Still believe that a callable with a set of pre-defined strings would be more convenient

SergeyTsimfer (approved):

batchflow/models/torch/base.py
HollowPrincess: Ok

batchflow/models/torch/base.py
```diff
@@ -789,6 +795,59 @@ def head(cls, inputs, **kwargs):
...
+
+        if self.model and ((init_model_weights is not None) or init_zero_bias):
+            # Parse model weights initilaization
+            if isinstance(init_model_weights, str):
+                # We have only one variant of predefined init function, so we check that init is str for a typo case
+                # The common used non-default weights initialization:
+                init_model_weights = common_used_weights_init
+
+            # Weights and biases initialization
+            for module in self.model.modules():
```
SergeyTsimfer, line 844: That should be a part of the default initialization

batchflow/models/torch/base.py
```diff
@@ -789,6 +795,59 @@ def head(cls, inputs, **kwargs):
...
+        --------
+            .. code-block:: python
+
+                def callable_init(module): # example of a callable for init
+                    if isinstance(module, nn.Linear):
+                        nn.kaiming_normal_(module.weight)
+
+                config = {'init_model_weights': callable_init}
+        """
+        def common_used_weights_init(module):
```
SergeyTsimfer, line 824: Move it to a separate `initialization.py` file, add the source of this init function

batchflow/models/torch/base.py
SergeyTsimfer, line 844: self.model.apply(function)

PR #630, opened 2021-12-14
Torch improvements

SergeyTsimfer:
This PR proposes to:

## Refactor the inputs/targets logic of `TorchModel`. 
Previously, we allowed to use `*args`, `**kwargs` and even `feed_dict` in both `train` and `predict` methods, as well as in the visualization tools like `get_gradcam`. This was ugly in code (and required ~50+ lines of parsing), and was prone to errors, especially in the multi-input or multi-output settings. It is very unintuitive to use as you can essentially pass anything to a model and hope it would arrange elements in correct order. Moreover, as `targets` have no defined argument, they are parsed by using hardcoded names, which is very confusing when works incorrectly.

In this PR, a different logic is implemented. All of the above methods now accept `inputs` and `targets` arguments, that are passed directly to the model or loss computation function.
Inside the code, they are always treated like sequences of `np.narrays`, which makes it easy to, for example, split both inputs and targets into microbatches, compute their shapes, and assert some additional properties for better user experience.
The only thing we (possibly) do with those arguments is, if there is only one element, we pass it to model / loss directly, so you don't need to worry about single-element sequences inside your `forward` pass.

## Improve the `fetches` behavior
The `fetches` parameter is renamed to `outputs`.
As previously, one can use pre-defined aliases to get parts of the network, i.e. `predictions` and `loss`. The list of aliases is extended by:
- `predictions_{i}` to get the i-th tensor of model output
- `predictions_{i}_{j}` to get the j-th operation, defined in the `outputs` config section, applied to the i-th tensor. For example, if the `config['outputs'] = {'first_tensor': ['sigmoid', 'softmax', custom_callable]}`, then the `predictions_0_1` gets the result of `softmax` operation, applied to the first tensor

Moreover, now we accept layer identifiers to get the intermediate activations from a neural network. The layer identifier is a string that defines how to get a layer in the model, and uses the same semantics, as the usual `getattr`/`getitem` way.

For example, if you can access a particular layer by 
```python
batchflow_model.model.body.encoder["block-1"][0].layer
```
then you can pass the 
```python
'model.body.encoder["block-1"][0].layer'
```
string as one of the `outputs` to retrieve the activations from exactly this part of the NN. Note that all of the layer identifiers start with `'model.'` prefix.

This logic is currently used only in `predict` method, not in `train`. Comment, if you see the use case for it to be in `train` as well.

## Re-arrange parts of the code to be more cohesive
The `TorchModel` code serves two main purposes: to create a `PyTorch model` instance and to provide infrastructure for its training. Now there is a clearer separation between the two.

Also, the `build` stage of the `TorchModel` instance is now explicitly split into `update instance attributes from passed config` and `update config with parsed attributes` stages. 

## Improve textual and graphical visualizations
- fixed a lot of newlines in the `.information` visualization
- added a loss plot
- more info is saved to `last_iteration_info` dictionary during the `train` pass
- a lot of methods are now public, so they have documentation and explanation of inner workings. Also, methods are reoranged to better correspond to logic of their usage

# Proposals
Additionally, I think that it would be beneficial to:

- Remove the `BaseModel` or at least move all the loading logic from it to the `TorchModel` itself. 
Now, as there is only one framework we support (#629), it would be easier to comprehend if all the code regarding `TorchModel` would be in the same place.

- Rename `.config` and `.full_config` attributes into `.external_config` and `.config` respectively. This would make it easier to understand the meaning of all of these configs.

- Make a separate `model_config`, which is subset of the main configuration by `order` and its elements keys.
As of now, there is no clear explanation to why we need to first parse instance attributes from config and then update config with some of the instance attributes. In reality, we need to update only the `model_config`, which makes sense, as model creation parameters can be unknown before we get additional information about the runtime (for example, we can infer shapes from passed batch data, which are unknown at instance initialization from config).

Update 15/12/2021: added `microbatch` to `predict`, as well as the `layer id` parsing logic to train.

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -930,7 +930,7 @@ class TorchModel(BaseModel, ExtractionMixin, VisualizationMixin):
         # Lock the entire method; release in any case
         try:
             if lock:
-                self.model_lock.acquire() #pylint: disable=consider-using-with
+                self.model_lock.acquire()
             self.last_iteration_info = {}
 
             # Parse inputs and targets: always a list
@@ -1215,7 +1215,7 @@ class TorchModel(BaseModel, ExtractionMixin, VisualizationMixin):
         # Acquire lock; release in any case
         try:
             if lock:
-                self.model_lock.acquire() #pylint: disable=consider-using-with
+                self.model_lock.acquire()
 
             # Parse outputs: always a list
             single_output = isinstance(outputs, str)
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -105,10 +105,10 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
         Keys are `inputs_shapes`, `targets_shapes`, `classes`, and `placeholder_batch_size`.
         By default, no shapes are set in the config.
 
-        - train and inference methods have common parameters:
+        - train and inference common parameters:
             - `amp` turns on/off automatic mixed precision, which allows to perform some of the operations in `float16`.
             Default is True.
-            - `microbatch` allows to split the training/inference batches in chunks (microbatches) and process
+            - `microbatch_size` allows to split the training/inference batches in chunks (microbatches) and process
             them sequentially. During train, we apply gradients only after all microbatches from the batch are used.
             Default is to not use microbatching.
 
@@ -231,12 +231,11 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
         Whether to use automated mixed precision during model training and inference. Default is True.
         The output type of predictions remains float32. Can be changed in `train` and `predict` arguments.
 
-    microbatch : int, bool or None
+    microbatch_size : int, bool or None
         Also known as virtual batch. Allows to process given data sequentially,
         accumulating gradients from microbatches and applying them once in the end.
         If int, then size of chunks to split every batch into.
-        If True, then every batch is split into individual items (same as microbatch equals 1).
-        If False or None, then feature is not used. Default is not to use microbatching.
+        If False or None, then this feature is not used. Default is not to use microbatching.
         Can be changed in `train` and `predict` arguments.
 
 
@@ -356,14 +355,14 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
         'loss': 'bdice',                                               # binary dice coefficient as loss function
         'optimizer': {'name': 'Adam', 'lr': 0.01,},
         'decay': {'name': 'exp', 'gamma': 0.9, 'frequency': 100},
-        'microbatch': 16,                                              # size of microbatches at training
+        'microbatch_size': 16,                                         # size of microbatches at training
     }
     """
     PRESERVE = [
         'full_config', 'config', 'model',
         'inputs_shapes', 'targets_shapes', 'classes',
         'loss', 'optimizer', 'decay', 'decay_step',
-        'sync_counter', 'microbatch',
+        'sync_counter', 'microbatch_size',
         'iteration', 'last_train_info', 'last_predict_info',
         'lr_list', 'syncs', 'decay_iters',
         '_loss_list', 'loss_list',
@@ -402,7 +401,7 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
         # Memory amortization: accumulate gradients to update weights later
         self.sync_frequency = 1
         self.sync_counter = 0
-        self.microbatch = None
+        self.microbatch_size = None
 
         # Sharpness-aware minimization
         self.sam_rho = 0.0
@@ -481,7 +480,7 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
             'amp': True,
             'device': None,
             'benchmark': True,
-            'microbatch': None,
+            'microbatch_size': False,
             'sync_frequency': 1,
             'profile': False,
 
@@ -522,7 +521,7 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
         config = self.config
 
         self.init_weights = config.get('init_weights', None)
-        self.microbatch = config.get('microbatch', None)
+        self.microbatch_size = config.get('microbatch_size', config.get('microbatch', False))
         self.sync_frequency = config.get('sync_frequency', 1)
         self.amp = config.get('amp', True)
 
@@ -944,7 +943,7 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
 
     # Apply model to train/predict on given data
     def train(self, inputs, targets, outputs=None, lock=True, profile=False,
-              sync_frequency=True, microbatch=True, microbatch_drop_last=True,
+              sync_frequency=True, microbatch_size=None, microbatch_drop_last=True,
               sam_rho=None, sam_individual_norm=None):
         """ Train the model with the data provided
 
@@ -968,11 +967,11 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
             If True, then value from config is used.
             Default value is 1, which means to apply gradients after each batch of data.
             If False or None, then gradients are applied after each batch of data.
-        microbatch : int, bool or None
+        microbatch_size : int, bool or None
             If int, then size of chunks to split every batch into. Allows to process given data sequentially,
             accumulating gradients from microbatches and applying them once in the end.
-            If True, then value from config is used (default value is not to use microbatching).
-            If False or None, then microbatching is not used.
+            If None, then value from config is used (default value is not to use microbatching).
+            If False, then microbatching is not used.
         microbatch_drop_last : bool
             Whether to drop microbatches, that are smaller than the microbatch size. Default is True.
         sam_rho : float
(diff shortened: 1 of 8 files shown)


batchflow/models/torch/base.py
```diff
@@ -736,44 +653,116 @@ def make_decay(self, decay, optimizer=None, **kwargs):
...
+        The order is defined by `order` key in the config, which is [`initial_block`, `body`, `head`] by default.
+        Each item in `order` should describe the block name, the config name and method to create. It can be a:
+            - string, then we use at as name, key and method name
+            - tuple of three elements, which are name, key and method name or callable
+            - dictionary with three items, which are `block_name`, `config_name` and `method`.
+
+        The `block_name` is then used as the identifier in resulting model, i.e. `model.initial_block`, `model.body`.
+        The `config_name` is used to retrieve block creation parameters from config.
+        The `method` is either a string to getattr from the current instance or a callable.
+        Either method or callable should return an instance of nn.Module and accept block parameters.
```
a-arefina, line 679: indent this block to match the dictionary description above

Last line is not quite clear: what do you refer to by 'method'

a-arefina: "name of the method to get from the current instance" - this description from below is much more clear

batchflow/models/torch/base.py
```diff
@@ -736,44 +653,116 @@ def make_decay(self, decay, optimizer=None, **kwargs):
...
+                block_name = config_name = method = item
+            elif isinstance(item, tuple) and len(item) == 3:
+                block_name, config_name, method = item
+            elif isinstance(item, dict):
+                block_name = item['block_name']
+                config_name = item.get('config_name', block_name)
+                method = item.get('method', config_name)
+
+            # Make block, from the `inputs`, transfer it to device
+            # Important: apply to the `inputs` before showing to the next block, so the shapes/etc are updated
```
a-arefina, line 699:
```suggestion
            # Important: apply to the `inputs` before passing them to the next block, so the shapes/etc are updated
```

batchflow/models/torch/base.py
```diff
@@ -736,44 +653,116 @@ def make_decay(self, decay, optimizer=None, **kwargs):
...
+        block = config[name]
+
+        if isinstance(block, nn.Module):
+            # Already initialized module
+            pass
+
+        elif isinstance(block, dict):
+            block_params = {**config['common'], **block}
+
+            if 'module' in block_params:
```
a-arefina, line 726: we also have `base` and `base_block`

batchflow/models/torch/base.py
```diff
@@ -863,63 +835,67 @@ def transfer_to_device(self, data):
...
-
-        output = []
-        for name in _fetches:
-            if name in outputs:
-                value = outputs[name]
-                value = self.transfer_from_device(value)
-                output.append(value)
-            else:
-                raise KeyError('Unknown value to fetch', name)
+        raise TypeError(f'Passed data should either be a `torch.Tensor` or a container of them, got {data}')
```
a-arefina, line 856: numpy arrays and sclalars are also allowed

batchflow/models/torch/base.py
```diff
@@ -863,63 +835,67 @@ def transfer_to_device(self, data):
...
+            Desired outputs of the method.
+            Each string defines a tensor to get and should be one of pre-defined or set in `outputs` key in the config.
+            Pre-defined tensors are `predictions`, `loss`, and `predictions_{i}` for multi-output models.
         lock : bool
             If True, then model, loss and gradient update operations are locked, thus allowing for multithreading.
         sync_frequency : int, bool or None
             If int, then how often to apply accumulated gradients to the weights.
-            If True, then value from config is used (default value is to apply gradients after each batch of data).
+            If True, then value from config is used. Default value is 1,
+            which means to apply gradients after each batch of data.
```
a-arefina, line 890:
```suggestion
            If True, then value from config is used. 
            Default value is 1, which means to apply gradients after each batch of data.
```

batchflow/models/torch/base.py
```diff
@@ -1200,174 +1210,128 @@ def predict(self, *args, targets=None, feed_dict=None, fetches=None, lock=True,
...
+            # For each operation, add multiple aliases
+            for j, operation in enumerate(output_operations):
+                output_tensor, operation_name = self.apply_output_operation(tensor, operation)
+                if operation_name:
+                    outputs[output_prefix + operation_name] = output_tensor # i.e. `first_sigmoid`, `sigmoid`
 
-            prefix = [*ops.keys()][i]
-            attr_prefix = prefix + '_' if prefix else ''
+                outputs.update({
+                    output_prefix + str(j) : output_tensor, # i.e. `first_0`, `0`
```
a-arefina, line 1304: so this key is overitten for multiple predictions?

batchflow/models/torch/visualization.py
```diff
@@ -97,6 +109,17 @@ def save_graph(self, log_dir=None, **kwargs):
         writer.add_graph(self.model, self._placeholder_data())
         writer.close()
 
+    def show_loss(self):
```
a-arefina, line 112: add yscale=log option

SergeyTsimfer:
Added `microbatch` to `predict`, as well as the `layer id` parsing logic to train.

batchflow/models/torch/base.py
```diff
@@ -1167,94 +1151,195 @@ def _train_sam_update_gradients(self, inputs, targets, sync_frequency, sam_rho,
...
+        >>> batchflow_model.predict(inputs=B.images, outputs='model.body.encoder["block-0"]')
         """
         # Acquire lock; release in any case
         try:
             if lock:
                 self.model_lock.acquire()
+            self.last_predict_info = {}
+
+            # Parse inputs and targets: always a list
+            inputs = list(inputs) if isinstance(inputs, (tuple, list)) else [inputs]
```
a-arefina, line 1199: to_list?

https://github.com/analysiscenter/batchflow/blob/7d201bfdf220496e966e9db5eb27a57212638994/batchflow/utils.py#L19

batchflow/models/torch/base.py
```diff
@@ -1167,94 +1151,195 @@ def _train_sam_update_gradients(self, inputs, targets, sync_frequency, sam_rho,
...
+
+    # Common utilities for train and predict
+    def split_into_microbatches(self, inputs, targets, microbatch, drop_last):
+        """ Split inputs and targets into microbatch-sized chunks. """
+        # Parse microbatch size
+        if microbatch:
+            if microbatch is True:
+                microbatch = self.microbatch
+            else:
+                microbatch = microbatch or self.microbatch
```
a-arefina, line 1291: Since `if microbatch` check above has passed, `or self.microbatch` will never be used

batchflow/models/torch/base.py
```diff
@@ -1167,94 +1151,195 @@ def _train_sam_update_gradients(self, inputs, targets, sync_frequency, sam_rho,
...
+
+        # Compute batch_size and make sure it is the same for all inputs and targets
+        batch_size = len(inputs[0])
+        for i, item in enumerate(inputs):
+            if len(item) != batch_size:
+                raise ValueError('All of `inputs` should have the same length, as the first one!'
+                                    f'Input at position `{i}` has size {len(item)}!={batch_size}')
```
a-arefina, line 1298:
```suggestion
                raise ValueError('All of `inputs` should have the same batch size!'
                                    f'Input at position `{i}` has batch size {len(item)}!={batch_size}')
```

batchflow/models/torch/base.py
```diff
@@ -1167,94 +1151,195 @@ def _train_sam_update_gradients(self, inputs, targets, sync_frequency, sam_rho,
...
+        for i, item in enumerate(targets):
+            if len(item) != batch_size:
+                raise ValueError('All of `targets` should have the same length, as the first of `inputs`!'
+                                    f'Target at position `{i}` has size {len(item)}!={batch_size}')
```
a-arefina, line 1302:
```suggestion
                raise ValueError('All of `targets` should have the same batch size as `inputs`!'
                                    f'Target at position `{i}` has batch size {len(item)}!={batch_size}')
```

batchflow/models/torch/base.py
```diff
@@ -1167,94 +1151,195 @@ def _train_sam_update_gradients(self, inputs, targets, sync_frequency, sam_rho,
...
+            chunked_targets = [targets]
+
+        return chunked_inputs, chunked_targets, batch_size, microbatch
+
+    def aggregate_microbatches(self, outputs, chunked_outputs, single_output):
+        """ Aggregate outputs from microbatches into outputs for the whole batch.
+        Scalar values are aggregated by `mean`, array values are concatenated along the first (batch) axis.
+        """
+        result = []
+        for i, _ in enumerate(outputs):
```
a-arefina, line 1325:
```suggestion
        for i in range(len(outputs)):
```

batchflow/models/torch/base.py
SergeyTsimfer: That is the functionality of `ConvBlock`

batchflow/models/torch/base.py
SergeyTsimfer: I really don't like the implementation of it, as it breaks in some cases (although not really relevant here until we try to pass anything but `np.ndarray`s)

Also, it would change one clear pythonic line to just another line of code with a function call, which may do unexpected things (and in the case it indeed would do that)

batchflow/models/torch/base.py
SergeyTsimfer: Nice catch

batchflow/models/torch/base.py
```diff
@@ -117,6 +207,62 @@ def callable_init(module): # example of a callable for init
...
+
+    # Shapes: optional
+    inputs_shapes : sequence
+        Shapes of the input tensors without the batch size.
+        Must be a tuple (one input) or sequence on tuples (multiple inputs) with shapes.
+
+    inputs_shapes : sequence
+        Shapes of the target tensors without the batch size.
+        Must be a tuple (one target) or sequence on tuples (multiple targets) with shapes.
+        Available as `` parameter in the `head` block.
```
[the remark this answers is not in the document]
HollowPrincess: And in other places

batchflow/models/torch/base.py
```diff
@@ -736,44 +729,117 @@ def make_decay(self, decay, optimizer=None, **kwargs):
...
+        After it, create training infrastructure (loss, optimizer, decay).
+
+        The order is defined by `order` key in the config, which is [`initial_block`, `body`, `head`] by default.
+        Each item in `order` should describe the block name, the config name and method to create. It can be a:
+            - string, then we use it as name, config key and method name
+            - tuple of three elements, which are name, config key and method name or callable
+            - dictionary with three items, which are `block_name`, `config_name` and `method`.
 
-    # Define model structure
+            The `block_name` is then used as the identifier in resulting model, i.e. `model.body`, `model.head`.
```
HollowPrincess, line 752:
```suggestion
            The `block_name` is used as the identifier in resulting model, i.e. `model.body`, `model.head`.
```

PR #631, opened 2021-12-17
Update notebook utils

SergeyTsimfer:
In this PR:

- change requests to jupyter lab API to a newer version
- add `use_pylintrc` option to `pylint_notebook` function: this allows to use the BatchFlow pylint configuration in the linted notebook as well

## Update: now the code supports both v2 and v3 `JupyterLab`

### DIFF
--- a/batchflow/utils_notebook.py
+++ b/batchflow/utils_notebook.py
@@ -35,19 +35,25 @@ def get_notebook_path():
 
     import requests
     import ipykernel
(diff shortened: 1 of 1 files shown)


a-arefina (commented):
I'm not quite sure that it is worth doing, but you can detect the library version that is critical for the new functionality and use, say, `notebook_utils_old.py` for older containers for the sake of backward compability

PR #633, opened 2021-12-28
Incremental improvements

SergeyTsimfer:
In this PR:

- A better `show_graphs` is implemeted for `TorchModel`, which shows both `loss` and `learning_rate` on the same axis. An example of created image:
image

- a `Pipeline.__str__` is implemented to show all of the actions and their parameters

- `Notifier` is improved in a sense of working with more APIs (now can be used as a context manager), not failing to work with iterables with no defined `__len__`, and extracting data from more containers (now can work with callables and mutable sequences)


batchflow/models/torch/base.py
```diff
@@ -1416,22 +1421,30 @@ def compute_outputs(self, predictions):
 
         outputs = {}
         # Iterate over tensors in predictions and the corresponding output operations
-        for i, (tensor, (output_prefix, output_operations)) in enumerate(zip(predictions, self.operations.items())):
+        iterator = zip_longest(predictions, self.operations.items(), fillvalue=(None, None))
```
a-arefina, line 1424: what if `predictions` is shorter than `self.operations.items()`? You'll have `(None, None)` and get an error in line 1426

batchflow/models/torch/mixins.py
```diff
@@ -140,6 +132,85 @@ def show_lr(self):
...
+    def show_loss(self, overlay_lr=True, figsize=(12, 6), window=20, final_window=50, minor_tick_frequency=4,
+                  log_loss=False, log_lr=False, return_figure=False, savepath=None, save_kwargs=None):
+        """ Plot loss and learning rate over the same figure.
+
+        Parameters
+        figsize : tuple of ints
+            Size of the figure.
+        window : int or None
+            If int, then averaged graph of loss values is displayed over the regular one.
+            The average for each point is computed as the mean value of the kernel with the center in this point.
```
a-arefina, line 144: Why center, not previous points? THis will skew the plot at the edges.

batchflow/pipeline.py
```diff
@@ -935,6 +976,15 @@ def m(self, name, batch=None):
...
         return self.get_model_by_name(name, batch=batch)
 
+    @property
+    def model(self):
+        """ An alias for a present model, if pipeline has only one model initialized. """
+        n_models = len(self.models.models)
+        if n_models == 1:
+            return list(self.models.models.values())[0]
+        raise TypeError('`model` property should be used only for pipelines with exactly 1 initialized model, '
+                        f'have {n_models} instead!')
```
a-arefina, line 986: You need some subject (grammatical) in the last phrase.

batchflow/models/torch/base.py
SergeyTsimfer: That relies on `outputs` being at least somewhat correct. I will add this check, but I can't save the user of this class from shooting off his/her leg

batchflow/models/torch/mixins.py
SergeyTsimfer: The moving average will be always somewhat skewed, to one side or to the other. Using the previous points makes the `average_loss` lag noticeably, which looks unnatural for most actual loss functions. Current implementation makes edges a bit smoother and with less wiggles, which is not as bad.

PR #634, opened 2022-01-10
`run_notebook` upgrade

HollowPrincess:
`run_notebook` upgrades: created new input/output logic with shelve, added traceback extraction, and small prettifyings.


batchflow/utils_notebook.py
```diff
@@ -144,51 +173,51 @@ def run_notebook(path, nb_kwargs=None, insert_pos=1, kernel_name=None, timeout=-
...
+                shelve.Pickler = Pickler
+                shelve.Unpickler = Unpickler
+
+                with shelve.open('{nb_kwargs_path}') as nb_locals:
+                    nb_kwargs = {{**nb_locals}}
+                    print("nb_kwargs =", nb_kwargs)
+
+                    locals().update(nb_kwargs)"""
+
+        code = re.sub(r' {8}', '', code) # drop extra spaces, that are needed for pretty code string writing
```
SergeyTsimfer, line 200: why not `dedent` wrom `textwrap`?

batchflow/utils_notebook.py
```diff
@@ -134,66 +167,120 @@ def run_notebook(path, nb_kwargs=None, insert_pos=1, kernel_name=None, timeout=-
...
+
+        if failed:
+            exec_res.update({'failed cell number': error_cell_num, 'traceback': traceback_message})
+
         # Add execution info
         if add_timestamp:
-            duration = int(time.time() - start_time)
             timestamp = (f'**Executed:** {time.ctime(start_time)}<br>'
-                         f'**Duration:** {duration} seconds.<br>'
+                         f'**Duration:** {int(time.time() - start_time)} seconds.<br>'
```
SergeyTsimfer, line 283: Actually, I would improve this by changing format to represent time in hours/minutes/seconds

batchflow/utils_notebook.py
```diff
@@ -134,66 +167,120 @@ def run_notebook(path, nb_kwargs=None, insert_pos=1, kernel_name=None, timeout=-
...
+            nb_outputs = [nb_outputs]
+
+        code = f"""
+                # Output dict preparation
+                output = {{}}
+                for value_name in {nb_outputs}:
+                    if value_name in locals():
+                        output[value_name] = locals()[value_name]
+                    else:
+                        output[value_name] = "Executed notebook doesn't contain this variable."
```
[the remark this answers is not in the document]
HollowPrincess: What if some variable has a `None` value? Maybe it will be better has no return of absent values?

SergeyTsimfer (commented):
Look at the function `run_notebook` as a whole and carefully think about its (possible) refactoring. Which parameters are really needed and useful and which are redundant? Which parts of the function should be moved to separate functions or should it be kept as a giant 200+lines monolith? How can we re-name / rethink parts of the functions to keep the logic of similar components aligned?

This function is not used widely as of now, so we can change it with to backwards compatibility, so let's use this chance before moving it to `nbtools`

batchflow/utils_notebook.py
```diff
@@ -86,6 +106,8 @@ def run_notebook(path, nb_kwargs=None, insert_pos=1, kernel_name=None, timeout=-
     nb_kwargs : dict, optional
         Additional arguments for notebook execution. Converted into cell of variable assignments and inserted
         into the notebook on `insert_pos` place.
+    nb_outputs : str or iterable of str
+        List of notebook locals names that return to output.
```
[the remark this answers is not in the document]
HollowPrincess: But these are not all local variables. Also, naming `local variables` doesn't provide information that these variables will be returned from the method.

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+        It is a required argument if inputs or outputs are provided.
+    execute_kwargs : dict, optional
+        Other parameters of `:class:ExecutePreprocessor`.
+    out_path_ipynb : str, optional
+        Path to save the output .ipynb file.
+    out_path_html : str, optional
+        Path to save the output .html file.
+    add_timestamp : bool
+        Whether to add a cell with execution information at the beginning of the executed notebook.
+    hide_input : bool
```
SergeyTsimfer, line 43: collides with `inputs` parameter -- are these about the same thing or not?

SergeyTsimfer: `hide_code_cells`

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+        It provides next information:
+            - 'failed' : bool
+               Whether the execution was failed.
+            - 'outputs' : dict
+               The notebook saved outputs.
+            - 'failed cell number': int
+              An error cell execution number (if exists).
+            - 'traceback': str
+              Traceback message from the notebook (if exists).
+    notebook :class:`nbformat.notebooknode.NotebookNode`
```
SergeyTsimfer, line 65: just add it to the dictionary, no need to return a tuple

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+    import nbformat
+    from jupyter_client.manager import KernelManager
+    from nbconvert.preprocessors import ExecutePreprocessor
+    from nbconvert import HTMLExporter
+    import shelve
+    from dill import Pickler, Unpickler
+    from textwrap import dedent
+
+    if not os.path.exists(path):
+        raise FileNotFoundError(f'Path {path} not found.')
```
SergeyTsimfer, line 80: That will be raised by `nbformat`

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+
+    if (inputs or outputs) and out_path_db is None:
```
SergeyTsimfer, line 82: Unlike other path-parameters, I think that this one should have a default (output_path + postfix)

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+    # Notebook preparation:
+    # Read the notebook, insert cell with inputs, create a notebook for outputs extraction
+    notebook = nbformat.read(path, as_version=4)
+
+    if hide_input:
+        notebook["metadata"].update({"hide_input": True})
+
+    if inputs or outputs:
+        # (Re)create a shelve database
+        shelve.Pickler = Pickler
```
SergeyTsimfer, line 105: That is pretty much a part of `import` process: move it there

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+
+                for value_name in outputs:
+                    if value_name in locals():
+                        output[value_name] = locals()[value_name]
+
+                with shelve.open(out_path_db) as notebook_db:
+                    notebook_db['outputs'] = output"""
+
+        code = dedent(code)
+        code = code_header + code
```
SergeyTsimfer, line 157: code = (code_header if not inputs else '') + code

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+
+        output_notebook = nbformat.v4.new_notebook(metadata=notebook.metadata)
+        output_notebook['cells'].append(nbformat.v4.new_code_cell(code))
+
+    # Execute the notebook
+    start_time = time.time()
+    try:
+        executor.preprocess(notebook, {'metadata': {'path': working_dir}}, km=kernel_manager)
+        exec_failed = False
```
SergeyTsimfer, line 166: `execution_failed` flag should be created (with `False` value) before `try/except`

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+                    if value_name in locals():
+                        output[value_name] = locals()[value_name]
+
+                with shelve.open(out_path_db) as notebook_db:
+                    notebook_db['outputs'] = output"""
+
+        code = dedent(code)
+        code = code_header + code
+
+        output_notebook = nbformat.v4.new_notebook(metadata=notebook.metadata)
```
SergeyTsimfer, line 159: Why is that a new notebook?

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+    except:
+        exec_failed = True
+        if raise_exception:
+            raise
+    finally:
+        # Save notebook outputs in the shelve db
+        if outputs is not None:
+            executor.preprocess(output_notebook, {'metadata': {'path': working_dir}}, km=kernel_manager)
+
+        # Check that something gone wrong or not
```
SergeyTsimfer, line 176: Check if something went wrong

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+        failed, error_cell_num, traceback_message = extract_traceback(notebook=notebook)
+        failed = failed or exec_failed
+
+        # Prepare execution results: execution state, notebook outputs and error info (if exists)
+        exec_res = {'failed': failed}
+        if failed:
```
SergeyTsimfer, line 182: you can put all three keys into the resulting dictionary regardless

batchflow/run_notebook.py
```diff
@@ -0,0 +1,253 @@
...
+
+        if add_timestamp:
+            timestamp = (f"**Executed:** {time.ctime(start_time)}<br>"
+                         f"**Duration:** {time.strftime('%H:%M:%S', time.gmtime(time.time() - start_time))}<br>"
+                         f"**Autogenerated from:** [{path}]\n\n---")
+            timestamp_cell = nbformat.v4.new_markdown_cell(timestamp)
+            notebook['cells'].insert(0, timestamp_cell)
+
+        # Save the executed notebook/HTML to disk
+        if out_path_ipynb:
```
SergeyTsimfer, line 197: maybe `notebook_to_html` / `save_notebook` funcitons?

batchflow/run_notebook.py
```diff
@@ -0,0 +1,273 @@
...
+        if out_path_db and not failed:
+            for ext in ['bak', 'dat', 'dir']:
+                os.remove(out_path_db + '.' + ext)
+
+        if return_notebook:
+            exec_res['notebook'] = notebook
+        return exec_res
+
+# Save notebook functions
+def save_notebook(notebook, out_path_ipynb, display_links):
```
SergeyTsimfer, line 218: `display_link`, here and in the next function

batchflow/run_notebook.py
```diff
@@ -0,0 +1,273 @@
...
+    import shelve
+    from dill import Pickler, Unpickler
+    from textwrap import dedent
+
+    if inputs or outputs:
+        # Set `out_path_db` value
+        if out_path_db is None:
+            if out_path_ipynb:
+                out_path_db = os.path.splitext(out_path_ipynb)[0] + '_db'
+            elif out_path_html:
```
SergeyTsimfer, line 83: fallback only on `out_path_ipynb`

PR #640, opened 2022-02-01
Research imp

AlexeyKozhevin:
Add `redirect_stdout` parameter to redirect `print`'s from experiments into files.

### DIFF
--- a/batchflow/research/experiment.py
+++ b/batchflow/research/experiment.py
@@ -6,6 +6,7 @@ import sys
 from copy import copy, deepcopy
 import itertools
 import traceback
+import contextlib
 from collections import OrderedDict
 import json
 import time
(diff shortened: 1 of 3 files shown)


batchflow/research/experiment.py
```diff
@@ -754,31 +756,46 @@ def close_logger(self):
...
-                    self.exception = e
-                    ex_traceback = e.__traceback__
-                    msg = ''.join(traceback.format_exception(e.__class__, e, ex_traceback))
-                    self.logger.error(f"Fail '{name}' [{iteration}/{n_iters}]: Exception\n{msg}")
-                    if self.monitor:
-                        self.monitor.fail_item_execution(name, self, msg)
-            if self.is_failed and ((list(self.actions.keys())[-1] == name) or (not self.executor.finalize)):
-                self.is_alive = False
+        if self.redirect_stdout in [2, 3] and self.dump_results:
+            if self.redirect_stdout == 1:
```
SergeyTsimfer, line 760: You will never enter this condition, as you checked that `self.redirect_stdout in [2, 3]`

PR #642, opened 2022-02-11
ConvNext

SergeyTsimfer:
https://arxiv.org/pdf/2201.03545.pdf

Also small improvements to `Notifier`, `Monitor` and `TorchModel`.

### DIFF
--- a/batchflow/models/torch/base.py
+++ b/batchflow/models/torch/base.py
@@ -370,6 +370,9 @@ class TorchModel(BaseModel, ExtractionMixin, OptimalBatchSizeMixin, Visualizatio
     ]
 
     def __init__(self, config=None):
+        if isinstance(config, str):
+            config = {'load/path': config}
+
         self.full_config = Config(config)
         self.model_lock = Lock()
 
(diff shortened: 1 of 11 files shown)


batchflow/models/torch/base.py
```diff
@@ -370,6 +370,9 @@ def callable_init(module): # example of a callable for init
     ]
 
     def __init__(self, config=None):
+        if isinstance(config, str):
```
roman-kh, line 373: Too much is implied, isn't it?

batchflow/models/torch/layers/conv.py
```diff
@@ -28,22 +28,32 @@ def __init__(self, filters, kernel_size=3, strides=1, padding='same',
...
 
-        if isinstance(padding, tuple) and isinstance(padding[0], tuple):
-            args['padding'] = 0
-            self.padding = sum(padding, ())
+            if isinstance(padding, tuple) and isinstance(padding[0], tuple):
+                args['padding'] = 0
+                self.padding = sum(padding, ())
+            else:
+                args['padding'] = padding
+                self.padding = 0
```
HollowPrincess, line 39:
```suggestion
                self.padding = None
```

PR #644, opened 2022-02-21
Plot

cdtn:
- [x] Add `(x, y)` data format support in `'curve'` mode.
- [x] Allow lists of numbers in `'curve'` mode
- [x] Fix `ticks` set with empty list
- [x] Fix `load_attribute` original src dictionary mutations
- [x] Make `suptitle` priority higher than `suptitle_label`
- [x] Test `colorbar` position for various figsizes
- [x] Allow call with no data
- [x] Add `text_color` param
- [x] Fix `bad_values` parsing
- [x] Think on `filter_parameters` logic simplification
- [x] Allow shapes of future subplots in `make_or_parse_axis`
- [x] `add_legend` text centering
- [x] Ensure that `figsize` is inferred correctly when `ncols` > `len(data)`
- [x] Ensure correct processing of axes created outside
- [x] Maintain legend parameters after adding new handles
- [x] Extend `add_text` doc.
- [x] Make clear parameters indexing error.
- [x] Refactor figsize adjustment code.
- [x] Re-check tutorials.
- [x] Ensure dpi kwargs work on save
- [x] Add dilation and masking by str expressions to `Layer` data preprocessing.
- [x] Add `set_data` for `plot` and `loss` modes
- [x] `imshow` -> `image`


batchflow/plot.py
```diff
@@ -0,0 +1,830 @@
...
+            ax.set_facecolor(facecolor)
+
+        ax.set_axisbelow(ax_config.get('set_axisbelow', False))
+
+        if ax_config.get('disable_axes'):
+            ax.set_axis_off()
+        elif not ax.axison:
+            ax.set_axis_on()
+
+    def show(self):
```
SergeyTsimfer, line 477: What about adding `finalize` option/method somewhere to explicitly clear everything?

batchflow/plot.py
```diff
@@ -0,0 +1,830 @@
...
+        return self
+
+    def __call__(self, mode, **kwargs):
+        self.plot(mode=mode, **kwargs)
+
+    def _ipython_display_(self):
+        self.show()
+
+    def __repr__(self):
+        return repr(self.fig).replace('Figure', 'Batchflow Figure')
```
SergeyTsimfer, line 375: Amen

batchflow/plot.py
```diff
@@ -0,0 +1,830 @@
...
+                raise ValueError(f"Not enough axes provided — got ({len(axes)}) for {len(self.shapes)} subplots.")
+
+        return fig, axes, config
+
+    def plot(self, data=None, mode='imshow', combine=None, save=False, axes_idx=None, **kwargs):
+        """ TODO
+
+        Notes to self: Explain `abs_idx` and `rel_idx`.
+        """
+        if data is not None:
```
SergeyTsimfer, line 344: Why do you need two ways to prepare the data, one in init and one here?

SergeyTsimfer (commented):
Will review the rest later. More remarks:
- not sure about the ordering of methods
- maybe add `loss` mode explicitly, which works with loss/lr curves. better yet, just modify the `plot_loss` of `TorchModel`
- do you really need to store actual data in the instances?

batchflow/plot.py
```diff
@@ -0,0 +1,830 @@
...
+        return getattr(self.im, key)
+
+    def __repr__(self):
+        if self.im is None:
+            return super().__repr__()
+        return self.im.__repr__()
+
+
+class plot:
+    """ Plotting backend for matplotlib.
```
AlexeyKozhevin, line 93: Mention the ways to display data (`imshow`, `wiggle`...). The name `plot` might suggest that it's all about line plotting.

batchflow/plot.py
```diff
@@ -0,0 +1,830 @@
...
+    But there is a way to provide parameter to a plot method, even if it's not hard-coded.
+    One must use specific prefix for that.
+    Address docs of `imshow`, `wiggle`, `hist`, `curve` and `annotate_axis` for details.
+
+    This also allows one to pass arguments of the same name for different plotting steps.
+    E.g. `plt.set_title` and `plt.set_xlabel` both require `fontsize` argument.
+    Providing `{'fontsize': 30}` in kwargs will affect both title and x-axis labels.
+    To change parameter for title only, one can provide {'title_fontsize': 30}` instead.
+    """
+    def __init__(self, data=None, combine='overlay', mode='imshow', **kwargs):
```
AlexeyKozhevin, line 143: This docstring is unavailable from the outside: move the `__init__` explanation and parameters to the class docstring

batchflow/plot.py
```diff
@@ -0,0 +1,830 @@
...
+from matplotlib.colors import ColorConverter, ListedColormap, is_color_like
+from matplotlib.lines import Line2D
+from matplotlib.patches import Patch
+from mpl_toolkits import axes_grid1
+
+from .utils import to_list
+
+
+
+class LoopedList(list):
```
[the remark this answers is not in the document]
cdtn: One can't index chain object

batchflow/plot.py
```diff
@@ -0,0 +1,1228 @@
...
+        yticks : `plt.set_yticks`
+        tick : `plt.tick_params`
+        xlim : `plt.set_xlim`
+        ylim : `plt.set_ylim`
+        legend : `plt.legend`
+        grid : `plt.grid`
+
+    Overall idea
+    ------------
+    Simply provide data, plot mode and parameters to the `plot` initialization
```
SergeyTsimfer, line 216: > General idea is to display graphs for provided data while passing other keyword arguments to correct `matplotlib` function, e.g. `title` goes to `plt.set_title`, `figsize` goes to figure initialization, etc.

batchflow/plot.py
```diff
@@ -0,0 +1,1228 @@
...
+        - Calculate subplots sizes - look through all data items
+          and estimate every subplot shape taking max of all its layers shapes.
+        - Put provided arrays into double nested list.
+          Nestedness levels define subplot and layer data order correspondingly.
+        - Infer images combination mode.
+        - Calculate indices corrections for empty subplots.
+    2. Parse figure axes if provided, else create them with either parsed parameters or inferred ones.
+    3. Obtain default config for chosen mode and merge them with provided config.
+    4. For every axis-data pair:
+        - If no data provided for axis, set if off.
```
SergeyTsimfer, line 230: set it off

batchflow/plot.py
```diff
@@ -0,0 +1,1228 @@
...
+        self.axes_objects = []
+
+        self.plot(data=data, combine=combine, mode=mode, **kwargs)
+
+    @staticmethod
+    def parse_data(data, combine, mode):
+        """ TODO """
+        contains_numbers = lambda x: isinstance(x[0], Number)
+
+        def process_tuple(data):
```
SergeyTsimfer, line 278: Why are you defining those functions inside this method?

batchflow/plot.py
```diff
@@ -0,0 +1,1228 @@
...
+        if y_n is not None:
+            set_locator = getattr(ax.yaxis, f'set_{grid_type}_locator')
+            set_locator(locator(y_n))
+
+        ax.grid(which=grid_type, zorder=zorder, **kwargs)
+
+
+
+def plot_image(data, **kwargs):
+    """ Shorthand for image plotting. """
```
SergeyTsimfer, line 1215: I would prefer to make some effort to keep the docstring of the class in this function. In the simplest case, you can use `partial` or explicitly set the `__doc__` of this function.

In more advanced, you could remove all of the sections corresponding to other modes and keep only the relevant part of documentation

PR #645, opened 2022-05-06
Torch improvements

SergeyTsimfer:
This PR refactors `TorchModel` with following goals in mind:

- decouple the process of model (as an instance of `torch.nn.Module`) creation and its usage (for train and predict).
- split `EncoderDecoder` functionality across multiple modules, which can be combined without relying on monolithic structure of the current implementation, and, generally, introduce a better defined logic of model building.
- add easier ways to import modules from other libraries, than the one currently available.
- improve textual `repr` of created models.

After this PR, the model structure is as follows:
- A separate class, inherited from `nn.Module`, to represent the whole instance of a model: `Network`.
- `Network` builds out of modules: `DefaultModule` (which does the same, as `initial_block/body/head` did before), `EncoderModule`, `DecoderModule`, etc. The type of module to use is defined by `type` key in the `config`.
- Each `Module` builds out of large blocks: `Block`, `Downsample`, `Upsample`, `ResBlock`, `ConvNextBlock`.
- Each `Block` is either inherited from `Block`, or uses it/smaller layers to build named elements of well-known nets (like ResBlock for ResNet).
- Default `Block` uses `MultiLayer`, which chains `layout`-defined sequence of layers into one sequence of operations.

The non-exhaustive list of changes:
- [x] refactored the logic of model creation: explicit `Network` class, explicit `*Module` and `*Block` classes.
- [x] introduced `LayerReprMixin` and `ModuleDictReprMixin` to make beatiful and customized textual representation of the model.
- [x] added a module to import models from `HuggingFace`.
- [ ] added a module to import models from `PaddlePaddle`.
- [ ] added a module to import models from `SemanticSegmentation`.
- [x] added a module to apply `torchvision` transforms as part of the model.
- [x] replace some of the custom layers/functions with its counterparts from `torchvision`: `center_crop`, for example.
- [x] changed the order of `[tensor, skip]` in `MultiLayer`
- [x] unify parameters of `batchflow` layers with ones from `PyTorch`: `units -> features, filters -> channels, strides -> stride`
- [x] `layers/wrapper_letters.py` is used to encapsulate the code, which needs circular imports to correctly function. To avoid errors, all of the questionable import statements are performed during class initialization.
- [x] changed `padding` logic of `Conv` layers.
- [x] refactored `Combine` layer with new initialization logic and `drop_path` operation.
- [x] added `DenseAlongAxis` layer, available as letter `F`.
- [x] rewrite docstrings from scratch.
- [x] refactor named models to account for the changes.
- [x] rewrite tests and remove redundant ones.
- [x] add README to explain the overall structure of code.
- [x] add a guide of updating the code to a newer version of `TorchModel`.

Open questions:
- names for `MultiLayer`, `Network`
- maybe, remove `Block` altogether? The only functionality it provides is to repeat/chain similar blocks, which is rarely used (`ResBlock`, `DenseBlock`) and can be implemented on other levels
- should we convert all of the named models to configs

Updates:
- [x] `channels_last` config option to convert both model and tensors
- [x] `trainable` config option to freeze some model parts for training
- [x] `DefaultModule` is capable of not applying its `forward` on `inference`
- [x] `k` / `K` letters for `shared kernel convolution` and `multi kernel convolution`
- [x] `q` / `Q` letter for conv / transposed conv, initialized to the weights of average pooling / bilinear upsampling
- [x] `skip` dictionary in `EncoderModule` to allow additional operations for skip connections
- [x] `model weights averaging`: `SWA` / `EMA` during model train
- [x] `wrap_tta` for applying test-time augmentations
- [x] `wrap_trt` for converting PyTorch model to TensorRT engine, getting ~3-5x speedup
- [x] `MultiLayer` now saves detailed information during initialization
- [x] replaced custom `crop` with `torchvision.transforms.funcitonal.center_crop`
- [x] refactored `padding` parameter in convolutions and poolings

Later TODOs:
- `SegFormerBlock` implementation
- Loaders from `PaddlePaddle` and `SemanticSegmentation`
- unify configuration of default parameters: currently, in `TorchModel` it is defined in `default_config` classmethod, in `*Module`s in `DEFAULTS` class attribute, in `*Block`s in initialization parameters.


batchflow/models/torch/blocks/core.py
```diff
@@ -0,0 +1,135 @@
...
+        self.base_block, self.n_repeats = base_block, n_repeats
+        self.args, self.kwargs = args, kwargs
+
+        self.shapes = {}
+        self.initialize(inputs, base_block, n_repeats, *args, **kwargs)
+
+
+    def initialize(self, inputs, base_block, n_repeats, *args, **kwargs):
+        """ !!. """
+        for r in range(n_repeats):
```
HollowPrincess, line 57: Rename `r`, please

batchflow/models/torch/layers/combine.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def mul_forward(self, inputs):
+        """ Multiplication with broadcasting. """
+        result = 1
+        for item in inputs:
+            result = result * item
+        return result
+
+    def mean_forward(self, inputs):
+        return torch.mean(inputs)
```
HollowPrincess, line 136: What is the `inputs` dtype - tensor or list? `torch.mean` input must be a tensor.

batchflow/models/torch/layers/core.py
```diff
@@ -49,24 +49,59 @@ class Dense(nn.Module):
     keep_dims : int, optional
         Dimensions to keep while flattening input, see :class:`~.Flatten`, by default 0.
     """
-    def __init__(self, units, bias=True, inputs=None, flatten=True, keep_dims=0):
+    def __init__(self, features, bias=True, inputs=None, flatten=True, keep_dims=0):
```
HollowPrincess, line 52: If we try namings like in [torch](https://pytorch.org/docs/stable/generated/torch.sum.html) why not `keepdim` instead of `keep_dims`?

batchflow/models/torch/layers/core.py
```diff
@@ -156,22 +191,21 @@ def forward(self, x):
...
 
                     batch_size = x.shape[0]
                     sizes = [round(batch_size*item) for item in self.multisample[:-1]]
                     residual = batch_size - sum(sizes)
                     sizes += [residual]
                 else:
-                    raise ValueError('Elements of multisample must be either all ints or floats,\
-                                      got instead {}'.format(self.multisample))
+                    raise ValueError(f'Elements of multisample must be either all ints or floats,\
+                                       got "{self.multisample}" instead!')
```
HollowPrincess, line 203:
```suggestion
                    raise ValueError('Elements of multisample must be either all ints or floats,'
                                       f' got "{self.multisample}" instead!')
```

will be prettier (without printed extra spaces)
image

batchflow/models/torch/modules/encoder_decoder.py
```diff
@@ -0,0 +1,282 @@
...
+                output_shapes = get_shape(tensor)
+
+                self[block_name] = block
+                self.shapes[block_name] = (input_shapes, output_shapes)
+
+
+    def forward(self, inputs):
+        inputs = inputs if isinstance(inputs, list) else [inputs]
+        tensor = inputs[-1]
+        i = 0
```
HollowPrincess, line 181: As I see, we didn't use this counter. What about to remove it?

batchflow/models/torch/modules/loaders.py
```diff
@@ -0,0 +1,95 @@
...
+            model = list(model.named_children())[0][1]
+        elif isinstance(attribute_name, str):
+            attribute_list = attribute_name.split('.')
+            for attr in attribute_list:
+                model = getattr(model, attr)
+
+        # Remove some of the modules
+        try:
+            model.layernorm = nn.Identity()
+            model.layer_norm = nn.Identity()
```
HollowPrincess, line 57: Why? Maybe this can be an optional behavior?

batchflow/models/torch/layers/multilayer.py
```diff
@@ -0,0 +1,269 @@
...
+        layer_args = layer_args.get(layer_name, {})
+        args = unpack_args(args, *counters)
+        args = {**args, **layer_args}
+        return args
+
+    def _make_modules(self, inputs):
+        """ Create necessary modules from instance parameters. """
+        self.shapes = {}
+        self.layout = self.layout or ''
+        self.layout = self.layout.replace(' ', '')
```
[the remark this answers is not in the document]
SergeyTsimfer: Why? It is the default for `layout` already

batchflow/models/torch/modules/encoder_decoder.py
```diff
@@ -0,0 +1,282 @@
...
+
+        self.shapes = {}
+        self.initialize(inputs, **kwargs)
+
+    def initialize(self, inputs, **kwargs):
+        """ !!. """
+        inputs = inputs[self.input_index] if isinstance(inputs, list) else inputs
+
+        # Parse parameters
+        num_stages = kwargs.pop('num_stages')
```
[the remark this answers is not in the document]
SergeyTsimfer: It is included in defaults of the module

PR #651, opened 2022-06-23
Plot improvements

cdtn:


batchflow/monitor.py
```diff
@@ -307,53 +320,23 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
             monitor.fetch()
             monitor.stop()
 
+    def plot(self, plotter=None, positions=None, savepath=None, **kwargs):
```
SergeyTsimfer, line 323: Keep the `visualize` alias

SergeyTsimfer (changes requested):

PR #652, opened 2022-06-26
refactoring code for better vectorization

maldil:
Hi, 

Thank you very much for your excellent work in analysiscenter/batchflow.
I am a graduate student at the University of Colorado-Boulder, studying the best practices of evolving ML codes. From our research, one of the most common evolution best practice in ML code is the migration of loop-based computations to vectorization, since this usually improves performance. We made the following changes in batchflow, which remove the FOR loop and use NumPy APIs. I carefully checked the modification to ensure that it does not break the code. I will gladly contribute. Please help me to merge this.

### DIFF
--- a/batchflow/models/torch/layers/combine.py
+++ b/batchflow/models/torch/layers/combine.py
@@ -2,7 +2,7 @@
 import torch
 from torch import nn
 from torchvision.transforms.functional import center_crop
-
+import numpy as np
 from ..utils import get_shape, get_num_channels, get_num_dims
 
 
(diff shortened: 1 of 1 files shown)


SergeyTsimfer (commented):
Hello,

Thanks for interest to our work and your contribution.

However, the `sum_forward` expects to work with a list of `torch.Tensors` and uses python built-in `sum` to invoke the `__add__` protocol, returning the tensor with result of summation after applying correct broadcasting rules. Therefore, the use of `sum` is very much intended, and not the `numpy.sum` or even `torch.sum` functions.

PR #665, opened 2022-08-02
Fixed resblock

Anastasiia-prog:


SergeyTsimfer (changes requested):
Don't create new files unless it is needed
Make your changes in the original `ResBlock` class -- otherwise, I can't see git diff and it is hard to compare
Also, this PR contains your previous (unmerged) PR code -- that is not how it should be

PR #666, opened 2022-08-03
resblock changed

Anastasiia-prog:


batchflow/models/torch/blocks/named_blocks.py
```diff
@@ -76,6 +76,7 @@ class ResBlock(Block):
     def __init__(self, inputs=None, layout='cnacn', channels='same', kernel_size=3, stride=1,
                  downsample=False, bottleneck=False, attention=None, groups=1, op='+a', branch=None,
                  n_reps=1, **kwargs):
+        width_mid_channels = 2 if groups > 1 else 1
```
HollowPrincess, line 79: The `width_mid_channels` seems like a not suitable name: it is not a `width`, it is a `factor`.
Also, don't use `mid` word reduction: `middle` is a short word and it is more clear.
And think about words order in the naming.

PR #669, opened 2022-08-26
Allow seed fix in `Sampler`

cdtn:


batchflow/sampler.py
```diff
@@ -638,3 +653,32 @@ def cart_prod(*arrs):
     """
     grids = np.meshgrid(*arrs, indexing='ij')
     return np.stack(grids, axis=-1).reshape(-1, len(arrs))
+
+def mix_samplers_seeds(left, right):
```
SergeyTsimfer, line 657: Should be a staticmethod of base `Sampler` class

batchflow/sampler.py
```diff
@@ -507,16 +524,15 @@ class ScipySampler(Sampler):
...
     """
     def __init__(self, name, seed=None, **kwargs):
-        super().__init__(name, seed, **kwargs)
+        super().__init__(name, seed=seed, **kwargs)
         name = _get_method_by_alias(name, 'ss')
         self.name = name
-        self.state = make_rng(seed)
         self.distr = getattr(ss, self.name)(**kwargs)
 
     def sample(self, size):
```
SergeyTsimfer, line 538: `rng` should be allowed to be passed as optional argument

batchflow/research/domain.py
```diff
@@ -254,7 +254,7 @@ def __init__(self, domain=None, **kwargs):
         self.n_updates = 0
         self.additional = True
         self.create_id_prefix = False
-        self.random_state = None
+        self.rng = None
```
SergeyTsimfer, line 257: There is a chain of `random_state` variables and their SeedSequences in `batchflow`:
```
---Pipeline
---Dataset
------Batch
---------inbatch_parallel workers (threads / processes / for-items)
```

And `Research` is even one level above that.

Have you read this entire chain of properties before changing it?

batchflow/sampler.py
SergeyTsimfer, line 657: You are mixing seeds, which is wrong. If I would mix two samplers in the proposed way and generate one random number, it would be the same, irrespectable of how many times each of those two samplers was called before creating the mixture.

You need to mix entropies, and there is a well-established (and used in other places in `batchflow`) way to do so: `np.random.SeedSequence`

```python
rng1, rng2
state1 = rng1.bit_generator.state['state']['state']
state2 = rng2.bit_generator.state['state']['state']
seed = np.random.SeedSequence([state1, state2])
rng = np.random.default_rng(seed)
```

While the difference between these two approaches is hard to come by in any realistic example, the latter is the official way to do so.

In either case, the current proposed way to seed the RNG in sampler would not work with `batchflow+seismiQB` ways to fix the randomization, and the only thing you need to actually fix the seed for `make_locations(sampler)` in `seismiQB` is the ability to pass custom `rng` into `Sampler.sample` call

SergeyTsimfer (changes requested):

PR #674, opened 2022-10-25
fixed torchmodel bug

AnnaAltynova:


batchflow/models/torch/base.py
```diff
@@ -1509,25 +1509,23 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
+                    output_ = np.concatenate([chunk_output[:chunk_size]
+                                              for chunk_output, chunk_size in zip(chunked_output, chunk_sizes)], axis=0)
                 else:
-                    output_ = torch.cat(chunked_output, dim=0)
+                    output_ = torch.cat([chunk_output[:chunk_size]
+                                         for chunk_output, chunk_size in zip(chunked_output, chunk_sizes)], dim=0)
                 result.append(output_)
             else:
                 result.append(np.mean(chunked_output))
-
```
SergeyTsimfer, line 1525: don't do that

most of the times, there is a reason for a certain amount of newlines

Issue #672, opened 2022-10-03
Refactor `inbatch_parallel`


SergeyTsimfer:
As the `inbatch_parallel` now not supposed to be used on its own, we can refactor it with following goals in mind:

- [ ] remove `_use_self` args
- [ ] remove `init`/`post` functions: the container with `init` should be passed directly from `Batch.apply_parallel`, and the results should be post-processed in the `Batch.apply_parallel` as well
- [ ] make `inbatch_parallel` a class: that would allow for easier introspection and parameter changes on the fly, for example, `target` to any other.

SergeyTsimfer:
use `inspect` to check whether `src/dst` should be passed to the method

roman-kh:
Avoid `inspect` if possible. Use `inbatch_parallel` arguments instead.

PR #677, opened 2022-11-07
Fix some bugs

cdtn:
1. Allow `BaseImageBatch` call with `src=None` (assuming that the `ix` itself is already a full path to image).
2. ~~Undo `src` and `dst` being passed by `BaseImagesBatch.apply_parallel` to batch actions (recent bug, introduced in 4a106c7228542a6bc5d69a66df8d86fab7d3b058).~~

### DIFF
--- a/batchflow/batch.py
+++ b/batchflow/batch.py
@@ -90,7 +90,7 @@ class Batch(metaclass=MethodsTransformingMeta):
 
     def __init__(self, index, dataset=None, pipeline=None, preloaded=None, copy=False, *args, **kwargs):
         _ = args
-        if  self.components is not None and not isinstance(self.components, tuple):
+        if self.components is not None and not isinstance(self.components, tuple):
             raise TypeError("components should be a tuple of strings with components names")
         self.index = index
         self._preloaded_lock = threading.Lock()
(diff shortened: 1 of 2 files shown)


batchflow/batch.py
```diff
@@ -634,7 +634,7 @@ def apply_parallel(self, func, init=None, post=None, src=None, dst=None, *args,
         # Compute result. Unbind the method to pass self explicitly
         parallel = inbatch_parallel(init=init, post=post, target=target, src=src, dst=dst)
         transform = parallel(type(self)._apply_once)
-        result = transform(self, *args, func=func, p=p, src=src, dst=dst,
+        result = transform(self, *args, func=func, p=p,
```
SergeyTsimfer, line 637: Nope
#656 fixes the same problem by adding `src/dst` to the signatures of methods.

In (hopefully, not too distant) future I will rework the decorator to be a class that can use `inspect` to look at the method signature, see #672

cdtn: Well, in that case all `ImagesBatch` methods need `src` and `dst` arguments. I guess it's a 2-minute fix. But why do you even need those parameters inside batch actions anyway?

SergeyTsimfer: This way, we can use `apply_parallel` decorator everywhere, instead of using `inbatch_parallel` somewhere (like it was previously)

PR #679, opened 2022-11-21
fixed 1d center_crop

AnnaAltynova:
minor fix to force resize so that `center_crop` will receive correct `output_size` in form `(H, W)` not only in 2D case

### DIFF
--- a/batchflow/models/torch/layers/combine.py
+++ b/batchflow/models/torch/layers/combine.py
@@ -1,9 +1,9 @@
 """ Layer to combine multiple inputs into one tensor. """
+import numpy as np
 import torch
 from torch import nn
-from torchvision.transforms.functional import center_crop
 
-from ..utils import get_shape, get_num_channels, get_num_dims
+from ..utils import get_shape, get_num_channels, get_num_dims, pad, center_crop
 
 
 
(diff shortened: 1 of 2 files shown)


batchflow/models/torch/layers/combine.py
```diff
@@ -115,8 +115,15 @@ def spatial_resize(self, inputs):
             shape = get_shape(item)
             dim = get_num_dims(item)
             spatial_shape = shape[-dim:]
-            if dim > 0 and spatial_shape != tuple([1]*dim) and spatial_shape != spatial_shape_:
-                item = center_crop(item, get_shape(inputs[0])[2:])
+            if dim > 0 and spatial_shape != tuple([1] * dim) and spatial_shape != spatial_shape_:
+                # Handle dim >= 2
```
AlexeyKozhevin, line 119: What about fix for 3D? The function will not make shapes equal, only last two axes

AnnaAltynova: updated

batchflow/models/torch/layers/combine.py
```diff
@@ -103,20 +103,24 @@ def extra_repr(self):
...
             spatial_shape = shape[-dim:]
-            if dim > 0 and spatial_shape != tuple([1]*dim) and spatial_shape != spatial_shape_:
-                item = center_crop(item, get_shape(inputs[0])[2:])
+            if dim > 0 and spatial_shape != tuple([1] * dim) and spatial_shape != target_shape:
+                if any(np.array(spatial_shape) < np.array(target_shape)):
+                    item = pad(item, spatial_shape, target_shape)
+                    if get_shape(item)[-dim:] == target_shape:
+                        resized.append(item)
+                        continue
+                item = center_crop(item, target_shape, dim)
```
nikita-klsh, line 123: What if we exploit the feature of `F.pad` to "unpad" given negative widths.
Then the single call encapuslates both `pad` and `center_crop`

```suggestion
                div, mod = np.divmod(np.array(target_shape) - spatial_shape, 2)
                ix = sum(zip(div, div+mod), ())
                item = F.pad(arr, ix[::-1])
```

PR #682, opened 2022-12-21
deformable conv

AnnaAltynova:
implemented deformable conv v3

### DIFF
--- a/batchflow/models/torch/layers/conv_complex.py
+++ b/batchflow/models/torch/layers/conv_complex.py
@@ -2,10 +2,9 @@
 from math import ceil, sqrt, prod
 
 import torch
+from torch import nn
 import torchvision.ops
 import torch.nn.functional as F
-from torch import nn
-from torch.nn.modules.utils import _pair
 
 from .conv import Conv
 from .combine import Combine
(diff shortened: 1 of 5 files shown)


HollowPrincess (changes requested):
I left little comments, PR looks OK.

batchflow/models/torch/layers/multilayer.py
```diff
@@ -87,7 +87,7 @@ class MultiLayer(ModuleDictReprMixin, nn.ModuleDict):
         'q': 'avg_pool_conv',
         'Q': 'bilinear_conv_transpose',
         'm': 'multi_scale_conv',
-
+        'y': 'deformable_conv2d',
```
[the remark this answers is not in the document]
AnnaAltynova: since we don't have much letters left, but any suggestions are welcome :)

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -3,6 +3,9 @@
 
 import torch
 from torch import nn
+import torch.nn.functional as F
+import torchvision.ops
+from torch.nn.modules.utils import _pair
```
SergeyTsimfer, line 8: I am strongly against using functions from other packages' `utils`, especially ones with underscore in its name: those functions can be changed drastically, and this particular function saves you like 10 characters at most with loss at readability.
Anyhow, there is already a `to_n_tuple` function in `batchflow.models.torch.utils`.

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -338,3 +341,84 @@ def __init__(self, inputs=None, kernel_size=(3, 5, 7), stride=1, dilation=1, gro
 
     def forward(self, x):
         return Combine(op=self.combine)([layer(x) for layer in self])
+
+    
+
+class DeformableConv2d(nn.Module):
+    """ Deformable convolution: apply convolution with learnable grid sampling locations and feature amplitudes modulation.
+    Grid offsets and modulation scalars are learned with separate convolution layers.
+    Wang, Wenhai, et al. "`InternImage: Exploring Large-Scale Vision Foundation Models with Deformable Convolutions
```
SergeyTsimfer, line 350: I would also add the link to v2 paper

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -338,3 +341,84 @@ def __init__(self, inputs=None, kernel_size=(3, 5, 7), stride=1, dilation=1, gro
...
+    <https://arxiv.org/abs/2211.05778>`_"
+
+    Parameters
+    ----------
+    offset_groups : int
+        `groups` for offset and modulation convolution layers. Default is 1.
+    """
+    LAYER = nn.Conv2d
```
SergeyTsimfer, line 358: As this layer is 2D only (and I doubt that `PyTorch` would add 3D implementation any soon), I would not use this class attribute and just hardcode `Conv2d` calls

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -338,3 +341,84 @@ def __init__(self, inputs=None, kernel_size=(3, 5, 7), stride=1, dilation=1, gro
...
+
+    def __init__(self, inputs=None, channels=None, kernel_size=3, stride=1, padding='same',
+                 offset_groups=1, groups=1, dilation=1, bias=False):
+        super().__init__()
+        dims = get_num_dims(inputs)
+        if dims != 2:
+            raise NotImplementedError("DeformableConv2d supports only 2d inputs.")
```
SergeyTsimfer, line 365: ...although I would keep this check

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -338,3 +341,84 @@ def __init__(self, inputs=None, kernel_size=(3, 5, 7), stride=1, dilation=1, gro
...
+            'groups': in_channels,
+            'kernel_size': kernel_size,
+            'stride': stride,
+            'dilation': dilation,
+            'bias': bias,
+        }
+
+        
+        self.args.update(compute_padding(padding=padding, shape=inputs.shape[-dims:], kernel_size=kernel_size,
+                                    dilation=dilation, transposed=False, stride=stride))
```
SergeyTsimfer, line 386: indent

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -338,3 +341,84 @@ def __init__(self, inputs=None, kernel_size=(3, 5, 7), stride=1, dilation=1, gro
...
+        self.args = {
+            'in_channels': in_channels, 
+            'out_channels': in_channels,
+            'groups': in_channels,
+            'kernel_size': kernel_size,
+            'stride': stride,
+            'dilation': dilation,
+            'bias': bias,
+        }
+
```
SergeyTsimfer, line 383: too many newlines

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -338,3 +341,84 @@ def __init__(self, inputs=None, kernel_size=(3, 5, 7), stride=1, dilation=1, gro
...
+        
+        self.args.update(compute_padding(padding=padding, shape=inputs.shape[-dims:], kernel_size=kernel_size,
```
SergeyTsimfer, line 385: why `[-dims:]`? It should be `[2:]`
While it is numerically the same in case of 2D inputs (4D tensors), it makes no sense otherwise

batchflow/models/torch/layers/conv_complex.py
```diff
@@ -338,3 +341,84 @@ def __init__(self, inputs=None, kernel_size=(3, 5, 7), stride=1, dilation=1, gro
...
+                                    dilation=dilation, transposed=False, stride=stride))
+        # Depthwise separable conv
+        self.depthwise_layer = self.LAYER(**self.args)
+        self.pointwise_layer = self.LAYER(in_channels=in_channels, out_channels=channels, kernel_size=(1, 1),
+                                           groups=groups, padding=(0, 0), stride=(1, 1), dilation=(1, 1))
+
+
+        # Offset
+        self.args.update({'out_channels': offset_groups * 2 * kernel_size[0] * kernel_size[1],
```
SergeyTsimfer, line 394: I dislike having the same `dict` updated and reused multiple times, especially when it is an instance attribute and supposed to reflect some properties of a layer. Make copies of parameter dicts for `offset` and `modulator` convs

SergeyTsimfer: or dont store it in instance at all. `padding`, `stride` and `dilation` can be retrieved from `depthwise_layer` the same way you get its `weight` and `bias`

SergeyTsimfer (commented):
I would like to have `v2` and `v3` versions. We can do it by having two classes or by having an `if-else` clause on `version` parameter. I do like the second option more, as it would result less code duplication

SergeyTsimfer (approved):

PR #684, opened 2023-01-11
fixed pooling in resblock

AnnaAltynova:


batchflow/models/torch/blocks/named_blocks.py
```diff
@@ -107,11 +107,15 @@ def __init__(self, inputs=None, layout='cnacn', channels='same', kernel_size=3,
...
+        if get_num_channels(inputs) != channels[-1] and np.prod(stride_downsample) != 1:
+            # If main flow changes the number of channels and resolution, then side branch applies stride in pooling.
             # No activation, because it will be applied after summation with the main flow
             branch_layout = branch.get('layout', 'vcn')
             branch_stride_type = 'pool_stride' if 'v' in branch_layout else 'stride'
+        elif get_num_channels(inputs) != channels[-1] and np.prod(stride_downsample) == 1:
+            # If main flow changes the number of channels and not resolution, so must do the side branch.
+            branch_layout = branch.get('layout', 'cn')
+            branch_stride_type = 'stride'
         else:
```
SergeyTsimfer, line 119: what if the number of channels is the same, but there is a downsampling in the main flow?

PR #699, opened 2023-05-24
Fix gpu monitor

AlexeyKozhevin:
Fix empty list in `gpu_list`

### DIFF
--- a/batchflow/monitor.py
+++ b/batchflow/monitor.py
@@ -234,7 +234,7 @@ class GPUResourceMonitor(ResourceMonitor):
         super().__init__(function=function, frequency=frequency, **kwargs)
 
         # Fallback to env variable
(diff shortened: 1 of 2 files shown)


batchflow/monitor.py
```diff
@@ -234,7 +234,7 @@ def __init__(self, function=None, frequency=0.1, gpu_list=None, **kwargs):
         super().__init__(function=function, frequency=frequency, **kwargs)
 
         # Fallback to env variable
-        if gpu_list is None:
+        if gpu_list is None or len(gpu_list) == 0:
```
SergeyTsimfer, line 237: `if not gpu_list`? seems like it is one of the cases where it is truly appropriate

SergeyTsimfer (approved):

PR #701, opened 2023-06-02
Change config

EvgeniyS99:
A new version of config supports any hashable object as key (like a simple dict):
image
However, due to the fact that tests were written for the old version of config which supports only ```str``` as a key, some of them are failed

### DIFF
--- a/batchflow/config.py
+++ b/batchflow/config.py
@@ -212,6 +212,7 @@ class Config(dict):
             A key in the dictionary. '/' is used to get value from nested dict.
         default : misc
             Default value if key doesn't exist in config.
+            If key has several variables, `default` can be a list with defaults for each variable.
 
         Returns
         -------


SergeyTsimfer (changes requested):
An overall impression is that not much has changed
Rewrite this thing from scratches with desired functionality (and tests) in mind

batchflow/config.py
```diff
@@ -23,45 +21,94 @@ def __init__(self, config=None, **kwargs):
...
+            parameters from kwargs also will be parsed and saved into self
         """
         if config is None:
-            self.config = Config.IAddDict()
-        elif isinstance(config, (dict, list)):
-            self.config = self.parse(config)
+            pass
         elif isinstance(config, Config):
-            self.config = config.config
+            super().__init__(config)
```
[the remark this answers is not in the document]
EvgeniyS99: No, it simpy returns the given config
image

batchflow/config.py
```diff
@@ -23,45 +21,94 @@ def __init__(self, config=None, **kwargs):
...
+            if np.any([len(item) != 2 for item in items]):
+                raise ValueError('tuples in list should represent pairs key-value'
+                                 ', and therefore must be always the length of 2')
+
+        for key, value in items:
+            if not isinstance(key, (str, Path)):
+                raise TypeError(f'only str and Path keys are supported, "{str(key)}" is of {type(key)} type')
+
+            if isinstance(key, str):
+                key = '/'.join(filter(None, key.split('/')))
```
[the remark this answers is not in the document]
EvgeniyS99: Agree, but what if we have three or more consecutive slashes? Also, I think that ```if``` clause is needed cause key could be another object, not necessarily a string

batchflow/config.py
```diff
@@ -78,14 +125,12 @@ def get(self, variables, default=None, config=None):
...
-        return val
+            value = self._get(variables, config=config, default=default)
+        return value
 
     def _get(self, variables, config=None, **kwargs):
-        if config is None:
-            config = self.config
         pop = kwargs.get('pop', False)
         has_default = 'default' in kwargs
         default = kwargs.get('default')
```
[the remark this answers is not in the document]
EvgeniyS99: Sorry, but I can't understand why it is bad practices in this case

batchflow/config.py
```diff
@@ -1,289 +1,245 @@
...
+        """
+        value = self._get(key, pop=True, **kwargs)
+        return value
+    
+    def _get(self, key, **kwargs):
 
-    def _get(self, variables, config=None, **kwargs):
-        if config is None:
-            config = self.config
         pop = kwargs.get('pop', False)
```
HollowPrincess, line 125: Add `pop=False` into the signature

batchflow/config.py
```diff
@@ -1,289 +1,245 @@
...
         default = kwargs.get('default')
+        default = Config(default) if isinstance(default, dict) else default
+
+        method = super().get if not pop else super().pop
 
         unpack = False
-        if not isinstance(variables, (list, tuple)):
-            variables = list([variables])
+        if not isinstance(key, list):
+            key = list([key])
```
HollowPrincess, line 134: Why not `key = [key]`?

batchflow/config.py
```diff
@@ -1,289 +1,245 @@
...
+                    del parent[k]
+
+            else:
+                if variable in self:
+                    value = method(variable)
                 else:
                     if has_default:
-                        val = _config.get(var_name, default)
+                        return default
                     else:
```
HollowPrincess, line 159: extra `else` here.
Please, provide some comments, because this method is hard to read and understand.

batchflow/config.py
```diff
@@ -1,289 +1,245 @@
...
-        for key, value in items:
-            if isinstance(value, dict):
-                value = self.parse(value)
-            if not isinstance(key, str):
-                raise TypeError(f'only str keys are supported, "{str(key)}" is of {type(key)} type')
-            key = '/'.join(filter(None, key.split('/'))) #merge multiple consecutive slashes '/' to one
-            self.put(key, value, new_config)
-        return new_config
+            for key, value in kwargs.items():
+                self.put(key, value)
```
HollowPrincess, line 193:
```python
iterable = other if isinstance(other, dict) else kwargs
for key, value in iterable.items():
    self.put(key, value)
```

batchflow/config.py
```diff
@@ -1,289 +1,245 @@
...
+
+    def __getitem__(self, key):
+        value = self._get(key)
         return value
 
-    def get(self, variables, default=None, config=None):
-        """ Returns variables from config
+    def __setitem__(self, key, value):
+        if key in self:
+            self.pop(key)
```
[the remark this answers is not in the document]
EvgeniyS99: > Also, why we can't write the `put` method in the way when we needn't `pop` before `put`?

I think, we can't, because we want to update value if key is already in config. However, there are some cases when pop is necessary. 
For example: `config  = Config({'a/b': 1, 'a/c': 2})`
 When we want to set a new value for `key='a'`, `config['a'] = dict(b=0, d=3)`, we want to get
 `config = {'a': {'b': 0, 'd': 3}}`.

batchflow/config.py
```diff
@@ -1,289 +1,245 @@
...
         Parameters
         ----------
-        config : dict, Config or None
+        config : dict, Config, list, tuple or None
             an object to initialize Config
             if dict, all keys and values slashes will be parsed into nested structure of dicts
-            and the resulting dictionary will be saved into self.config
-            if an instance on Config, config.config will be saved to self.config (not a copy!)
+            and the resulting dictionary will be saved into self
+            if list or tuple, should contain key-value pairs with the length of 2
```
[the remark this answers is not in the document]
EvgeniyS99: > Do we really need this option?

Yes, this is used in __add__ method

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
+
+        return self
+
+    def put(self, key, value):
+        """ Put a new key into config recursively.
+
+        Parameters
+        ----------
+        key : hashable object
+            key to add. '/' is used to put value into nested dict
```
AlexeyKozhevin, line 72:
```suggestion
            Key to add. '/' is used to put value into nested dict
```

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
-    def _get(self, variables, config=None, **kwargs):
-        if config is None:
-            config = self.config
-        pop = kwargs.get('pop', False)
-        has_default = 'default' in kwargs
+        Parameters
+        ----------
+        key : str or list of hashable objects
+            '/' is used to get value from nested dict.
+        default : masc
```
AlexeyKozhevin, line 124:
```suggestion
        default : misc
```

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
+        return self
+
+    def put(self, key, value):
+        """ Put a new key into config recursively.
+
+        Parameters
+        ----------
+        key : hashable object
+            key to add. '/' is used to put value into nested dict
+        value : masc
```
AlexeyKozhevin, line 73:
```suggestion
        value : misc
```

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
+        """
+        if isinstance(key, str) and '/' in key:
+            config = self
+            parent, child = key.split('/', 1)
+
+            if parent in config and isinstance(config[parent], Config): # for example, we put value=3 with key='a/c' into the
+                config[parent].update(Config({child: value}))           # config = {'a': {'b': 1}} and want to receive {'a': {'b': 1, 'c': 3}}
+            else:
+                config[parent] = Config({child: value})
```
AlexeyKozhevin, line 82: We will silently rewrite items like `{'a': 3}` by `{'a/b': 2}`. Should it be a warning or exception?

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
-    class IAddDict(dict):
-        """ dict that supports update via += """
-        def __iadd__(self, other):
-            if isinstance(other, dict):
-                self.update(other)
-            else:
-                raise TypeError(f"unsupported operand type(s) for +=: 'IAddDict' and '{type(other)}'")
-            return self
 
+class Config(dict):
```
AlexeyKozhevin, line 3: Reorder methods in more structured way:
- init (`__init__`, `parse`)
- methods for elements (`put`, `get`, `pop`, `_get`)
- dict methods (`update`, `flatten`, `copy`, `keys`, `values`, `items`)
- content dunder methods (`__getitem__`, `__getitem__`, `__getattr__`, `__delitem__`)
- arithmetic dunder methods (`__iadd__`, ...)
- pickle dunder (`__getstate__`, `__setstate__`)

AlexeyKozhevin (changes requested):

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
 
-    def _get(self, variables, config=None, **kwargs):
-        if config is None:
-            config = self.config
-        pop = kwargs.get('pop', False)
-        has_default = 'default' in kwargs
+        Parameters
+        ----------
+        key : str or list of hashable objects
+            '/' is used to get value from nested dict.
```
HollowPrincess, line 123: This docstring line doesn't describe the `key` argument.

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
+        Single value or a tuple
+        """
+        value = self._get(key, pop=True, default=default)
+        return value
+
+    def _get(self, key, pop=False, **kwargs):
+        """ Recursively retrieve values for a given key if the key contains '/'.
+        If key doesn't contain '/', get or pop from the `dict` class is invoked.
+
+        Example:
```
HollowPrincess, line 139: https://numpydoc.readthedocs.io/en/latest/format.html#examples

batchflow/config.py
```diff
@@ -1,250 +1,257 @@
...
 
-    def parse(self, config):
-        """ Parses flatten config with slashes
+    def __getattr__(self, key):
+        if key in self:
+            value = self.get(key)
+            value = Config(value) if isinstance(value, dict) else value
+            return value
+        raise AttributeError(key)
+
```
HollowPrincess, line 207: Maybe `AttributeError` should be defined in the `self.get`?

batchflow/config.py
```diff
@@ -255,35 +262,11 @@ def __radd__(self, other):
...
-
-    def __len__(self):
-        return len(self.config)
-
+    def __eq__(self, other):
+        self_ = self.flatten() if isinstance(self, Config) else self
+        other_ = Config(other).flatten() if isinstance(other, dict) else other
+        return self_.__eq__(other_)
+    
     def __rshift__(self, other):
```
HollowPrincess, line 270: Please, fix docs here.

batchflow/config.py
EvgeniyS99: But in this case we will catch `AttributeError` every time we want to retrieve the key from the dict, not only by `.`

EvgeniyS99: > Maybe `AttributeError` should be defined in the `self.get`?

But in this case we will catch `AttributeError` every time we want to retrieve the key from the dict, not only by `.`

batchflow/config.py
EvgeniyS99: > We will silently rewrite items like `{'a': 3}` by `{'a/b': 2}`. Should it be a warning or exception?

I think, yes, a warning would be good here

batchflow/config.py
```diff
@@ -344,4 +289,66 @@ def items(self, flatten=False):
...
+    def __getitem__(self, key):
+        value = self._get(key)
+        return value
+
+    def __setitem__(self, key, value):
+        _ = self.pop(key, None)
+        self.put(key, value)
+
+    def __delitem__(self, key):
+        self.pop(key)
```
HollowPrincess, line 303: Why not:
```suggestion
    def __delitem__(self, key):
        _ = self.pop(key, None)
```
?

batchflow/config.py
```diff
@@ -344,4 +289,66 @@ def items(self, flatten=False):
...
+            value = Config(value) if isinstance(value, dict) else value
+            return value
+        raise AttributeError(key)
+    
+    def __iadd__(self, other):
+        if isinstance(other, dict):
+            self.update(other)
+        else:
+            raise TypeError(f"Unsupported operand type(s) for +=: 'Config' and '{type(other)}'")
+        return self
```
HollowPrincess, line 317:
```suggestion
    def __iadd__(self, other):
        if not isinstance(other, dict):
            raise TypeError(f"Unsupported operand type(s) for +=: 'Config' and '{type(other)}'")

        self.update(other)
        return self
```

batchflow/config.py
```diff
@@ -344,4 +289,66 @@ def items(self, flatten=False):
...
+
+    def __radd__(self, other):
+        if isinstance(other, dict):
+            other = Config(other)
+        return other.__add__(self)
+
+    def __eq__(self, other):
+        self_ = self.flatten() if isinstance(self, Config) else self
+        other_ = Config(other).flatten() if isinstance(other, dict) else other
+        return self_.__eq__(other_)
```
HollowPrincess, line 334: What if other is a `Config` and not `flatten`?

EvgeniyS99: > What if other is a `Config` and not `flatten`?

He will be flattened anyway because if other is an instance of `Config`, `isinstance(other, dict)` is True

batchflow/config.py
```diff
@@ -344,43 +288,111 @@ def values(self, flatten=False):
...
+
+            Returns
+            -------
+            Pipeline
+                Pipeline object with an updated config.
+        """
+        return other << self
+
+    def __getstate__(self):
+        """ Must be explicitly defined for pickling to work. """
```
SergeyTsimfer, line 393: Sounds very dogmatic

SergeyTsimfer (commented):
Looks a lot better!
A few comments:
- there are some dev comments in the code, but not enough of them in the `_get`/`_put` methods
- check that documentation is up-to-date
- add descriptive messages into raised errors: that would help greatly when keys are missing on various levels of nestedness

PR #704, opened 2023-06-27
Add Normalizer and Quantizer

AlexeyKozhevin:


batchflow/utils_transforms.py
```diff
@@ -0,0 +1,179 @@
...
+    mode : {'mean', 'std', 'meanstd', 'minmax'} or callable
+        If str, then normalization description.
+        If callable, then it will be called on `src` data with additional `normalization_stats` argument.
+    clip_to_quantiles : bool
+        Whether to clip the data to quantiles, specified by `q` parameter.
+        Quantile values are taken from `normalization_stats`, provided by either of the ways.
+    q : tuple of numbers
+        Quantiles for clipping. Used as keys to `normalization_stats`, provided by either of the ways.
+    normalization_stats : dict, optional
+        If provided, then used to get statistics for normalization.
```
SergeyTsimfer, line 21: `... Otherwise, compute them from a given array. `

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,179 @@
...
+                    np.clip(array, *np.quantile(array, self.q), out=array)
+                    clip_to_quantiles = False
+
+                if callable(self.mode):
+                    normalization_stats = {
+                        'mean': np.mean(array),
+                        'std': np.std(array),
+                        'min': np.min(array),
+                        'max': np.max(array),
+                    }
```
SergeyTsimfer, line 68: Do we even need those in case of callable `mode`? Maybe, pass an empty dict and let this callable mode do whatever

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,179 @@
...
+            Whether to return stats used for normalization, by default False
+
+        Returns
+        -------
+        numpy.ndarray or (numpy.ndarray, dict)
+        """
+        clip_to_quantiles = self.clip_to_quantiles
+        array = array if self.inplace else array.copy()
+
+        if normalization_stats is None:
```
SergeyTsimfer, line 54:
```python
normalization_stats = normalization_stats if normalization_stats is not None else self.normalization_stats

if normalization_stats is None:
    if clip:
        ...

    if callable:
        ...

    else:
        ...
```

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,179 @@
...
+                array /= normalization_stats['std'] + 1e-6
+            if 'min' in self.mode and 'max' in self.mode:
+                if clip_to_quantiles:
+                    array -= normalization_stats['q_01']
+                    array /= normalization_stats['q_99'] - normalization_stats['q_01']
+                elif normalization_stats['max'] != normalization_stats['min']:
+                    array -= normalization_stats['min']
+                    array /= normalization_stats['max'] - normalization_stats['min']
+                else:
+                    array -= normalization_stats['min']
```
SergeyTsimfer, line 98: What do you think about this way of parsing parts of string `mode` VS hardcoded options (`meanstd`, `minmax`, maybe `mean`)?

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,179 @@
...
+                array += normalization_stats['mean']
+            if 'min' in self.mode and 'max' in self.mode:
+                if normalization_stats['max'] != normalization_stats['min']:
+                    array *= normalization_stats['max'] - normalization_stats['min']
+                    array += normalization_stats['min']
+                else:
+                    array += normalization_stats['min']
+        return array
+
+class Quantizer:
```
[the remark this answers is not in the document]
AlexeyKozhevin: I think it's worth bringing it here to use with some kind of image. For example, this thing may be needed not only in interpretation, but also in processing

batchflow/utils_transforms.py
AlexeyKozhevin: I like the current scheme

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,217 @@
...
+                if 'mean' in mode:
+                    normalization_stats['mean'] = np.mean(array)
+                if 'std' in mode:
+                    normalization_stats['std'] = np.std(array)
+                if 'min' in mode:
+                    normalization_stats['min'] = np.min(array)
+                if 'max' in mode:
+                    normalization_stats['max'] = np.max(array)
+
+        if clip_to_quantiles:
```
SergeyTsimfer, line 76: That would clip twice in case of `normalization_stats is None`

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,217 @@
...
+
+        # Actual normalization
+        if callable(mode):
+            array[:] = mode(array, normalization_stats)
+        else:
+            if 'mean' in mode:
+                array -= normalization_stats['mean']
+            if 'std' in mode:
+                array /= normalization_stats['std'] + 1e-6
+            if 'min' in mode and 'max' in mode:
```
SergeyTsimfer, line 87: `if mode == 'minmax'`

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,217 @@
...
+        self.copy = copy
+
+        n_bins = np.iinfo(dtype).max - np.iinfo(dtype).min - 1
+        self.bins = np.histogram_bin_edges(None, bins=n_bins, range=ranges).astype(np.float32)
+
+    def quantize(self, array):
+        """ Quantize data: find the index of each element in the pre-computed bins and use it as the new value.
+        Converts `array` to `self.dtype`. Lossy.
+        """
+        if self.copy:
```
SergeyTsimfer, line 181: looks like `copy` should be a parameter of this function, not of the instance itself

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def compute_mean_error(self, data):
+        """ Estimate quantization error on data. """
+        quantized_data = self.quantize(data)
+        dequantized_data = self.dequantize(quantized_data)
+        return np.mean(np.abs(dequantized_data - data)) / data.std()
+
+    @property
+    def estimated_absolute_error(self):
+        return np.diff(self.bins).max()
```
SergeyTsimfer, line 214: Aren't bins uniform (meaning they have the same size) at this point?

batchflow/utils_transforms.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __call__(self, array):
```
SergeyTsimfer, line 216: Put it in the same relative position, as in `Normalizer`: right after the `normalize` method.

Also, you can write `__call__ = quantize` and that would save the documentation / signature / etc

batchflow/utils_transforms.py
AlexeyKozhevin: No 
image

SergeyTsimfer (approved):

SergeyTsimfer (approved):

PR #705, opened 2023-07-11
Added weighted BCE

EvgeniyS99:


AlexeyKozhevin (changes requested):

batchflow/models/torch/losses/binary.py
```diff
@@ -191,3 +193,27 @@ def forward(self, prediction, target):
...
+        mask[mask == 0] = num_positive / (num_positive + num_negative)
+
+        loss = torch.zeros(1, device=prediction.device)
+        loss += F.binary_cross_entropy_with_logits(prediction, target, weight=mask)
+
+        return loss
+
+losses = [WeightedBCE(), Dice()]
+weights = [.5, .5]
+comboloss = Weighted(losses, weights)
```
AlexeyKozhevin, line 219: Do you really need it here? There is not a problem to make it in your notebook:

ComboLoss = Weighted([WeightedBCE(), Dice()], [.5, .5])

Nevertheless, `comboloss` here is inappropriate naming.

AlexeyKozhevin (changes requested):

batchflow/models/torch/losses/binary.py
EvgeniyS99: I think, ye, but I thought I should present the full loss used in my models

EvgeniyS99: > Do you really need it here? There is not a problem to make it in your notebook:
> 
> ComboLoss = Weighted([WeightedBCE(), Dice()], [.5, .5])
> 
> Nevertheless, `comboloss` here is inappropriate naming.

Why is comboloss inappropriate naming? The bce loss is **combined** with dice loss --> **comboloss**

batchflow/models/torch/losses/binary.py
```diff
@@ -191,3 +191,23 @@ def forward(self, prediction, target):
         sensitivity = (squared_error * inverse).sum() / (inverse.sum() + self.eps)
 
         return self.r * specificity + (1 - self.r) * sensitivity
+
+
+
+class BalancedWeightedBCE(nn.Module):
```
SergeyTsimfer, line 197: `CrossEntropyLoss` from `batchflow.models.torch.losses.core.py` already can do all of that. Maybe, make inherit BCE / make it a special case of it somehow?

PR #708, opened 2023-07-25
Activations

EvgeniyS99:


batchflow/models/torch/utils.py
```diff
@@ -118,3 +120,48 @@ def center_crop(inputs, target_shape, dims):
     crops_ = [slice(crop_lefts[i], crop_lefts[i] + target_shape[i]) for i in range(dims)]
     crops = [slice(dim) for dim in no_crop_shape] + crops_
     return inputs[crops]
+
+
+def get_blocks_and_activations(model):
```
SergeyTsimfer, line 125: Too much hardcode here: if this code becomes a part of `TorchModel`, it should utilize and use understanding of classes involved. As of now, this code works only for a selected subclass of models.

As model (class `Network`) consists of pre-defined modules (`EncoderModule`, `DecoderModule`, `DefaultModule`, etc), you can check the instance of submodules and extract necessary names.

batchflow/models/torch/utils.py
```diff
@@ -118,3 +120,48 @@ def center_crop(inputs, target_shape, dims):
...
+    blocks += [f'model.decoder["{block}"]' for block in decoder_blocks]
+
+    activation_names = [f'encoder_{i}' for i in range(len(encoder_blocks))]
+    if embedding:
+        activation_names += ['embedding']
+    activation_names += [f'decoder_{i}' for i in range(len(decoder_blocks))]
+
+    return blocks, activation_names
+
+def compress_activations(batch, activation_names, **kwargs):
```
SergeyTsimfer, line 145: `batch` has no relation to `TorchModel` and will not be part of the model-related-utils

batchflow/models/torch/utils.py
```diff
@@ -118,3 +120,48 @@ def center_crop(inputs, target_shape, dims):
...
+        activation_images = getattr(batch, activation_name).copy()
+        if not np.isnan(activation_images.min()):
+            compressed_images, explained_variance = reduce_channels(activation_images, **kwargs)
+            setattr(batch, activation_name, compressed_images)
+        else:
+            return None
+
+    return batch, explained_variance
+
+def reduce_channels(images, n_components=3, **kwargs):
```
SergeyTsimfer, line 159: Would like an option to apply a user-provided `PCA` instance (so that we can accumulate data in it). Also, normalization should be optional

SergeyTsimfer (changes requested):
Don't mix different entities in code: each class/module should be responsible for its own thing. Rethink architecture of the proposed changes and their location in our code. One example of this can be:
- getting names / layers of a `TorchModel/Network` can be a part of the model files, maybe even as a method of it. In that case, it should be automatic and not hardcoded;
- an utility function to reduce the number of channels can be placed near to the model. Extra thought: you can do PCA on GPU: in that case, it can be integrated to the model `outputs` parameter smoothly and would probably look beautiful;
- any batch-related things look absolutely random here and do not belong. Do we even need them?

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+
+    # Utilities for activations
+    @staticmethod
+    def get_blocks_and_activations(model, modules=None):
+        """ Retrieve intermediate blocks of the neural network model
+        and corresponding activation names.
+
+        Parameters
+        ----------
+        model : Network
```
HollowPrincess, line 1863: Needs description or empty line after parameter declaration.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+        modules : str or list of str, default None
+            The main parts of the model for retrieving activations from.
```
HollowPrincess, line 1865: What main and non-main parts of the model are?)

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+            If None, all of the parts from the model.config['order'] wil be used.
```
HollowPrincess, line 1866:
```suggestion
            If None, all of the parts from the `model.config['order']` will be used.
```

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+
+        Returns
+        -------
+        activation_blocks : list of str
+            List with activation blocks which will be passed into `outputs` parameter of `self.predict`.
```
HollowPrincess, line 1871: Description must contain information only in context of the current object (method  `get_blocks_and_activations`).

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+        activation_names : list of str
+            List with names of activations corresponding to activation blocks.
```
HollowPrincess, line 1873: Please, try to simplify sentences everywhere, e.g.:
`List with names of activations` -> `Activation names list`

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+
+        Examples
+        --------
+        >>> get_blocks_and_activations(model, modules=['encoder', 'embedding', 'decoder'])
+        activation_blocks = ['model.encoder["block-0"]', 'model.encoder["block-1"]', 'model.embedding', 'model.decoder["block-0"]']
+        activation_names = ['encoder_0', 'encoder_1', 'embedding_0', 'decoder_0']
```
HollowPrincess, line 1879: I am not sure, that the output corresponds to the docstyle.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+                    if 'block' in block:
+                        extracted_blocks[module].append(f'model.{module}["{block}"]')
+            else:
+                extracted_blocks[module].append(f'model.{module}')
+
+        activation_blocks = []
+        activation_names = []
+        for module, blocks in extracted_blocks.items():
+            activation_blocks.extend(blocks)
+            activation_names += [f'{module}_{i}' for i in range(len(blocks))]
```
HollowPrincess, line 1897: Looks like this cycle is an extra one: try to do this in the previous for-loop.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+        return activation_blocks, activation_names
+
+    @staticmethod
+    def reduce_channels(pca_instance, images, normalize=True):
+        """ Convert multichannel 'b c h w' images to RGB images using PCA. """
+        images = images.transpose(0, 2, 3, 1)
+        compressed_images = pca_instance.fit_transform(images.reshape(-1, images.shape[-1]))
+        compressed_images = compressed_images.reshape(*images.shape[:3], pca_instance.n_components)
+        if normalize:
+            compressed_images = (compressed_images - compressed_images.min()) / (compressed_images.max() - compressed_images.min())
```
HollowPrincess, line 1908: Handle zero division error.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+        activation_names = []
+        for module, blocks in extracted_blocks.items():
+            activation_blocks.extend(blocks)
+            activation_names += [f'{module}_{i}' for i in range(len(blocks))]
+
+        return activation_blocks, activation_names
+
+    @staticmethod
+    def reduce_channels(pca_instance, images, normalize=True):
+        """ Convert multichannel 'b c h w' images to RGB images using PCA. """
```
HollowPrincess, line 1903: Think about arguments order.
Also, can we have a default `pca_instance` value for `pca_instance = None` ?
Information about images shape is more suitable in `Parameters` section.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1850,60 @@ def _parse_profilers(self):
...
+            else:
+                extracted_blocks[module].append(f'model.{module}')
+
+        activation_blocks = []
+        activation_names = []
+        for module, blocks in extracted_blocks.items():
+            activation_blocks.extend(blocks)
+            activation_names += [f'{module}_{i}' for i in range(len(blocks))]
+
+        return activation_blocks, activation_names
```
SergeyTsimfer, line 1899: Do you even need `activation_names`?

batchflow/models/torch/base.py
SergeyTsimfer: +1
`pca_instance` should be the last (and optional) parameter

batchflow/models/torch/base.py
EvgeniyS99: > Do you even need `activation_names`?

I use them to assign activations to batch, as well as to provide titles for plots.

batchflow/models/torch/base.py
```diff
@@ -33,6 +33,12 @@
 from ..base import BaseModel
 from ...config import Config
 
+from collections.abc import Iterable
+
+from ...utils_transforms import Normalizer
+
+from sklearn.decomposition import PCA
```
SergeyTsimfer, line 40: Pylint is unhappy

batchflow/models/torch/base.py
SergeyTsimfer: Well, it is how *you* use them. Why something related to your code (batch and plots), that don't belong here, should be there?

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1853,87 @@ def _parse_profilers(self):
                                          columns=['ncalls', 'CPU_tottime', 'CPU_cumtime', 'CUDA_cumtime'])
         self.profile_info['CPU_tottime_avg'] = self.profile_info['CPU_tottime'] / self.profile_info['ncalls']
         self.profile_info['CUDA_cumtime_avg'] = self.profile_info['CUDA_cumtime'] / self.profile_info['ncalls']
+
+
+    # Utilities for activations
+    @staticmethod
+    def get_blocks_and_activations(model, modules=None):
```
SergeyTsimfer, line 1860: That is not a staticmethod. The first argument is expected to be the model itself: essentially, `self`.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1853,87 @@ def _parse_profilers(self):
...
+        >>> print(activation_names)
+        ['encoder_0', 'encoder_1', 'embedding_0', 'decoder_0']
+        """
+        modules = [modules] if isinstance(modules, str) else modules or model.config['order']
+        activation_names = []
+        activation_blocks = []
+
+        for module in modules:
+            number = -1
+            extracted_module = getattr(model, module)
```
SergeyTsimfer, line 1893: Try to not mix naming schemes. Here, `module` is a `str`, but `extracted_module` is a `nn.Module`. Use `module_name` instead.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1853,87 @@ def _parse_profilers(self):
...
+            if isinstance(extracted_module, Iterable):
```
SergeyTsimfer, line 1894: I would rather check that `extracted_module` is an instance of `EncoderModule / DecoderModule / ...`: you are really trying to check that, not the fact that you can iterate over the module.

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1853,87 @@ def _parse_profilers(self):
...
+        compressed_images : np.ndarray
+            Compressed RGB images of shape (B, H, W, 3).
+        explained_variance_ratio : np.ndarray
+            Percentage of variance explained by each of the selected components.
+        """
+        images = images.transpose(0, 2, 3, 1)
+        if pca_instance is None:
+            pca_instance = PCA(n_components=3)
+
+        compressed_images = pca_instance.fit_transform(images.reshape(-1, images.shape[-1]))
```
SergeyTsimfer, line 1931: Is there a difference between re-creating instance each time or using passed one? Is it even possible to continuously `fit` PCA?

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1853,87 @@ def _parse_profilers(self):
...
+                        activation_blocks.append(f'model.{module}["{block}"]')
+                        activation_names.append(f'{module}_{number}')
+            else:
+                activation_blocks.append(f'model.{module}')
+                activation_names.append(f'{module}_0')
+
+        return activation_blocks, activation_names
+
+    @staticmethod
+    def reduce_channels(images, normalize=True, pca_instance=None):
```
SergeyTsimfer, line 1907: `images -> array`
`images -> tensor`
`images -> x`
Choose one

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1851,73 @@ def _parse_profilers(self):
                                          columns=['ncalls', 'CPU_tottime', 'CPU_cumtime', 'CUDA_cumtime'])
         self.profile_info['CPU_tottime_avg'] = self.profile_info['CPU_tottime'] / self.profile_info['ncalls']
         self.profile_info['CUDA_cumtime_avg'] = self.profile_info['CUDA_cumtime'] / self.profile_info['ncalls']
+
+
+    # Utilities for activations
+    def get_activation_blocks(self, modules=None):
+        """ Retrieve intermediate blocks of the neural network model
+        and corresponding activation names.
```
SergeyTsimfer, line 1859: Why not in one line?

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1851,73 @@ def _parse_profilers(self):
...
+            If None, all of the parts from the `model.config['order']` will be used.
+
+        Returns
+        -------
+        activation_blocks : list of str
+            Intermediate activation blocks.
+
+        Examples
+        --------
+        >>> model.get_blocks_and_activations(modules=['encoder', 'embedding', 'decoder'])
```
SergeyTsimfer, line 1874: old name

batchflow/models/torch/base.py
```diff
@@ -1847,3 +1851,73 @@ def _parse_profilers(self):
...
+        >>> print(activation_blocks)
+        ['model.encoder["block-0"]', 'model.encoder["block-1"]', 'model.embedding', 'model.decoder["block-0"]']
+        """
+        modules = [modules] if isinstance(modules, str) else modules or self.model.config['order']
+        activation_blocks = []
+
+        for module_name in modules:
+            extracted_module = getattr(self.model, module_name)
+            if isinstance(extracted_module, (EncoderModule, DecoderModule, MLPDecoderModule)):
+                for block in extracted_module:
```
SergeyTsimfer, line 1884: would rather use `block_name` naming

SergeyTsimfer (approved):

PR #710, opened 2023-07-27
added ADE dataset

BulatVakhitov:


batchflow/opensets/ade.py
```diff
@@ -0,0 +1,185 @@
...
+import tempfile
+
+from PIL import Image
+import tqdm
+import requests
+
+from . import ImagesOpenset
+
+
+class ADESegmentation(ImagesOpenset):
```
SergeyTsimfer, line 15: Please, add tests to the `tests` directory of `batchflow`:
- as part of the testing process, initialize different opensets and get one batch of its contents (images, masks, labels, whatever the dataset provides)
- mark tests as `slow`, so that they are not run at each commit
- make it in 100 lines of code or less

As an alternative idea, we can make an ipynb notebook that showcases datasets, but that is just an idea.

PR #711, opened 2023-07-27
Fix aggregate microbatches bug

BulatVakhitov:
chunked_output[0].size != 1 always true when passing torch.Tensor(size is a method). Changed it to get_size() function. Also the np.mean function falls, when passing torch tensors with device cuda, so I added .item() to all elements of chunked_output.


batchflow/models/torch/base.py
```diff
@@ -1551,7 +1551,7 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
                                          for chunk_output, chunk_size in zip(chunked_output, chunk_sizes)], dim=0)
                 result.append(output_)
             else:
-                result.append(np.mean(chunked_output))
+                result.append(np.mean([chunk_output.item() for chunk_output in chunked_output]))
```
SergeyTsimfer, line 1554: The only way we get here (in this method) an instance of `torch.Tensor` is when the data was explicitly left on device by setting `transfer_from_device=False`. It is rather strange to return a `np.scalar`, which is on CPU, in that case

SergeyTsimfer (approved):

PR #714, opened 2023-08-04
Added tests for opensets

BulatVakhitov:
Skipped test for COCOSegmentation because of the Response 406 during the mask download.

### DIFF
--- a/batchflow/tests/opensets_test.py
+++ b/batchflow/tests/opensets_test.py
@@ -1,0 +1,96 @@
+# pylint: disable=missing-docstring, redefined-outer-name
+
+import sys
+import pytest
+
+import PIL
+import numpy as np
+
+sys.path.insert(0, '../../')
+from batchflow.opensets import (PascalSegmentation, PascalClassification,
+                                COCOSegmentation, MNIST, CIFAR10, CIFAR100, Imagenette)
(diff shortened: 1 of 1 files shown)


batchflow/tests/opensets_test.py
```diff
@@ -0,0 +1,96 @@
...
+def imagenette():
+    return Imagenette()
+
+
+class TestOpensets:
+    @pytest.mark.slow
+    def test_pascal_segmentation(self, pascal_segmentation):
+        batch = pascal_segmentation.next_batch(batch_size=10)
+        assert isinstance(batch.images[0], PIL.JpegImagePlugin.JpegImageFile)
+        assert isinstance(batch.labels[0], PIL.PngImagePlugin.PngImageFile)
```
SergeyTsimfer, line 53: Tests look very similar, so you can write less functions and use `pytest.mark.parametrize`

SergeyTsimfer: Also, you don't need those fixtures as they are implemented now. There is no difference between them and instantiating datasets directly

SergeyTsimfer (commented):
Please, fix the failing tests

SergeyTsimfer (approved):

PR #716, opened 2023-08-17
Delayed and missing imports

SergeyTsimfer:
This PR adds a mechanism of delayed imports and applies it to slow and rarely used ones (`dask`, `feather`).

Also, it provides a way to postpone raising `ImportError` on missing imports, which is used for imports requiring extra dependencies.


batchflow/utils_import.py
```diff
@@ -0,0 +1,128 @@
...
+    def loaded_module(self):
+        """ Try loading the module at the first access. """
+        if self._loaded_module is None:
+            try:
+                self._loaded_module = import_module(self.module, self.package)
+                if self.attribute is not None:
+                    self._loaded_module = getattr(self._loaded_module, self.attribute)
+            except ImportError as e:
+                if self.help:
+                    raise ImportError(f"No module named '{self.module.strip()}'! {self.help}") from e
```
roman-kh, line 47: Do we really need `strip`? As it might conceal some typos in the module name which would be hard to detect.

roman-kh (approved):

batchflow/utils_import.py
```diff
@@ -70,7 +70,9 @@ def __getitem__(self, key):
 
 
 def make_delayed_import(module, package=None, attribute=None, help=None):
-    """ Make delayed import only if needed. """
+    """ Make delayed import only if needed.
+    Setting `BATCHFLOW_IMMEDIATE_IMPORT` variable to any value makes all imports immediate.
```
roman-kh, line 74: environment variable

PR #744, opened 2024-06-17
New way of finding optimal batch size

BulatVakhitov:
Now optimal batch_size can be found via running `compute_optimal_batch_size` function, which does two steps:

1.  Batch Size Estimation:
    - `predictive` method: This approach runs the model with various batch sizes.
    It then calculates the optimal batch size by solving a linear system of equations,
    `measured_memory = batch_size * item_size + model_size + eps`, to find both item_size and model_size.
    - `bruteforce` method: This approach runs the model with progressively larger batch sizes
    until an Out Of Memory (OOM) error occurs.

2. Optimal Batch Size Calculation:
    The exact optimal batch size is determined using a binary search algorithm.


batchflow/models/torch/base_mixins.py
```diff
@@ -221,6 +222,197 @@ class OptimalBatchSizeMixin:
...
+        )
+
+    def is_out_of_cpu_memory(self, exception):
+        return (
+            isinstance(exception, RuntimeError)
+            and len(exception.args) == 1
+            and "DefaultCPUAllocator: can't allocate memory" in exception.args[0]
+        )
+
+    def is_oom_error(self, exception):
```
SergeyTsimfer, line 250: The entire error parsing part can be simplified to just a few lines in one method

batchflow/models/torch/base_mixins.py
```diff
@@ -221,6 +222,197 @@ class OptimalBatchSizeMixin:
...
+        try:
+            # This is the last thing that should cause an OOM error, but seemingly it can.
+            torch.cuda.empty_cache()
+        except RuntimeError as exception:
+            if not self.is_oom_error(exception):
+                # Only handle OOM errors
+                raise
+
+
+    def _compute_optimal_batch_size(self, method='train', inputs=None, targets=None,
```
SergeyTsimfer, line 267: It is quite strange that this method is "private", while it uses "public" methods inside

SergeyTsimfer: should rename things into:
- `compute_optimal_batch_size`
- `_compute_batch_size_predictive`, maybe with no underscore
- `_compute_batch_size_bruteforce`, maybe with no underscore

Also, as there are now more methods, documentation should be moved from the Mixin to the specific methods

batchflow/models/torch/base_mixins.py
```diff
@@ -221,6 +222,197 @@ class OptimalBatchSizeMixin:
...
+                if self.is_oom_error(exception):
+                    high.append(batch_size[-1])
+                    batch_size.append((high[-1] + low[-1]) // 2)
+                else:
+                    raise # some other error not memory related
+
+            if high[-1] - low[-1] <= 1:
+                break
+
+            self.garbage_collection_cuda()
```
SergeyTsimfer, line 411: move this before the `if/break`: this way, you would not need to call `gc` after the for-loop

batchflow/models/torch/base_mixins.py
```diff
@@ -221,6 +222,197 @@ class OptimalBatchSizeMixin:
...
+                    batch_size.append(batch_size[-1] // factor)
+                    break
+                raise # some other error not memory related
+
+            self.garbage_collection_cuda()
+        self.garbage_collection_cuda()
+        return batch_size[-1], batch_size[-1] * factor
+
+
+    def _run_binary_scaling(self, inputs=None, targets=None, low=2, high=None,
```
SergeyTsimfer, line 348: function is pretty much identical to the previous one: only the way to compute the next tested batch size differs

batchflow/models/torch/base_mixins.py
```diff
@@ -221,6 +222,197 @@ class OptimalBatchSizeMixin:
...
+
+        low = low if isinstance(low, list) else [low]
+        high = high if isinstance(high, list) else [high]
+        batch_size = start_batch_size if isinstance(start_batch_size, list) else [start_batch_size]
+
+        n_iters = int(np.ceil(np.log2(high[-1] - low[-1])))
+        notifier = Notifier(n_iters=n_iters, bar=pbar,
+                            monitors=[{'source': low, 'name': 'low'},
+                                      {'source': high, 'name': 'high'},
+                                      {'source': batch_size, 'name': 'batch_size'}])
```
SergeyTsimfer, line 365: making all of the variables an instance of `list` so that Notifier can track them is not intuitive; while I like the idea, it should be commented in code

batchflow/models/torch/base_mixins.py
```diff
@@ -221,6 +222,197 @@ class OptimalBatchSizeMixin:
...
+            and len(exception.args) == 1
+            and "DefaultCPUAllocator: can't allocate memory" in exception.args[0]
+        )
+
+    def is_oom_error(self, exception):
+        return (self.is_cuda_out_of_memory(exception) or
+                self.is_cudnn_snafu(exception) or
+                self.is_out_of_cpu_memory(exception))
+
+    def garbage_collection_cuda(self):
```
SergeyTsimfer, line 255: hey, if you copy some things from other places, at least re-write them to your (and project) code-style

batchflow/models/torch/base_mixins.py
```diff
@@ -221,6 +222,197 @@ class OptimalBatchSizeMixin:
...
+    def _compute_optimal_batch_size(self, method='train', inputs=None, targets=None,
+                                    start_batch_size=4, max_iters=25, factor=2,
+                                    spread=0.2, use_estimation=False, n=4,
+                                    max_memory=100, pbar='n', tail_size=20,
+                                    frequency=0.05, delta_batch_size=16,
+                                    max_batch_size=1024, max_iters_estimation=2):
+        """ Compute memory usage for multiple batch sizes. """
+
+        # first calculate optimal batch_size estimation
+        if use_estimation:
```
SergeyTsimfer, line 276: would rather have parameter `estimation_method='predictive'/'bruteforce'`

SergeyTsimfer (commented):
- is it true that with `benchmark=False` the predictive optimal batch size computation can be run with much lower `n`, maybe even `n=1`?
- documentation is lacking, both in the master branch and the new implemented methods;
- overall looks good, just a few things to polish:)

batchflow/models/torch/base_mixins.py
```diff
@@ -221,10 +222,255 @@ class OptimalBatchSizeMixin:
...
+                raise
+
+
+    def compute_optimal_batch_size(self, method='train', inputs=None, targets=None,
+                                   start_batch_size=4, max_iters=25, factor=2,
+                                   spread=0.2, estimation_method='bruteforce', n=4,
+                                   max_memory=100, pbar='n', tail_size=20,
+                                   frequency=0.05, delta_batch_size=16,
+                                   max_batch_size=1024, max_iters_estimation=2):
+        """ Computes optimal batch size in two steps:
```
SergeyTsimfer, line 252: `Compute`

batchflow/models/torch/base_mixins.py
```diff
@@ -221,10 +222,255 @@ class OptimalBatchSizeMixin:
...
+        1. Calculate batch size estimation using `predictive` or `bruteforce` method.
+        2. Calculate exact optimal batch size using binary search.
+
+        Parameters
+        ----------
+        method: str
```
SergeyTsimfer, line 258: if the argument should be one of the predefined string values, document it as `method: str, {'option1', 'option2'}`

batchflow/models/torch/base_mixins.py
```diff
@@ -221,10 +222,255 @@ class OptimalBatchSizeMixin:
...
+            Defines in which method (`train` or `predict`) the optimal batch size will
+            be computed. Default: `train`.
+
+        estimation_method: str
+            Whether `bruteforce` or `predictive` estimation method will be used.
+
+        inputs: np.array
```
SergeyTsimfer, line 265: `np.ndarray`

batchflow/models/torch/base_mixins.py
```diff
@@ -221,10 +222,255 @@ class OptimalBatchSizeMixin:
...
+            The inputs to the model, that will be used in optimal batch computation.
+            If none, then the placeholder will be created
```
SergeyTsimfer, line 267: `If not provided`
Also, if you want to explicitly mention `None` value, capitalize the first letter

batchflow/models/torch/base_mixins.py
```diff
@@ -221,10 +222,255 @@ class OptimalBatchSizeMixin:
...
+
+        start_batch_size: int
+            Batch size to start batch_size estimation. If your model is small, you should use larger
+            start_batch_size in order to get accurate optimal batch_size
+
+        max_iters: int
+            Maximum number of binary search iterations.
+
+        factor: int
+            Value by which we multiply batch_size at each iteration. Uses in bruteforce estimation.
```
SergeyTsimfer, line 281: Use quotes to mention other parameters: \`batch_size\`

batchflow/models/torch/base_mixins.py
```diff
@@ -221,10 +222,255 @@ class OptimalBatchSizeMixin:
...
+
+        spread: int
+            Defines how wide the interval for binary search will be. Uses in predictive estimation.
```
SergeyTsimfer, line 284: This value is not an int; also, the documentation gives very little insight into how it works / what values should the user pass

batchflow/models/torch/base_mixins.py
```diff
@@ -221,10 +222,255 @@ class OptimalBatchSizeMixin:
...
+        max_batch_size: int
+            Maximum batch_size
+
+        max_iters_estimation: int
+            Maximum number of batch_size estimation iterations.
+
+        Returns
+        -------
+        optimal_batch_size: int
+            Batch size that perfectly fits in max_memory
```
SergeyTsimfer, line 315: UwU

SergeyTsimfer (commented):
I do like this version a lot more!
What do you think about moving thix mixin into a separate .py file?

batchflow/models/torch/batch_opt_mixin.py
```diff
@@ -0,0 +1,336 @@
+""" Contains mixin for :class:`~.torch.TorchModel` to provide textual and graphical visualizations. """
```
SergeyTsimfer, line 1: Its not

SergeyTsimfer: Also, I would rather name this file `base_batchsize_mixin.py`

SergeyTsimfer (approved):

batchflow/models/torch/batch_opt_mixin.py
BulatVakhitov: What's wrong with `batch_opt_mixin.py`? I think it's more self-explanatory

SergeyTsimfer: All of the mixins for `TorchModel` are now in `base_*.py` files: that is the main reason

BulatVakhitov: Fair enough

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,336 @@
...
+                                    start_batch_size=4, max_iters=15, pbar='n',
+                                    method='train', max_memory=100, frequency=0.01,
+                                    tail_size=20, update_method='bruteforce',
+                                    factor=2):
+        count = 0
+        # if None => make equal distance between low, start_batch_size and high
+        high = high if high is not None else 2 * start_batch_size - low
+
+        # The list is used to show current batch_size in notifier. Current value is the last one.
+        batch_size = start_batch_size if isinstance(start_batch_size, list) else [start_batch_size]
```
HollowPrincess, line 193: Maybe it would be better to separate these values into the two variables: `batch_size` and `batch_size_history`?

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,336 @@
...
+
+                # monitor consumed memory
+                with GPUMemoryMonitor(frequency=frequency) as monitor:
+                    input = inputs or self.make_placeholder_data(batch_size[-1], to_device=False)
+                    input = list(input) if isinstance(input, (tuple, list)) else [input]
+                    input = [item[:batch_size[-1]] for item in input]
+
+                    if method == 'train':
+                        target = targets or self.predict(inputs=input, outputs='predictions')
+                        target = list(target) if isinstance(target, (tuple, list)) else [target]
```
HollowPrincess, line 221: Hmmm, looks like in both cases it is a `list(target)`?

batchflow/models/torch/base_batchsize_mixin.py
HollowPrincess: It is better to use `list(start_batch_size)` instead of `[start_batch_size]`

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,336 @@
...
+                                               tail_size=tail_size, std_threshold=std_threshold)
+            table[batch_size] = info
+
+            # Exit condition
+            batch_size += delta_batch_size
+            if info['memory'] > max_memory or batch_size > max_batch_size:
+                break
+
+        # Make and solve a system of equations for `item_size`, `model_size`
+        matrix = np.array([[batch_size, 1] for batch_size in table.keys()])
```
HollowPrincess, line 281: Maybe it would be better to use preallocation with items assignment instead of comprehension?

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,336 @@
...
+        vector = np.array([value['memory'] for value in table.values()])
+        item_size, model_size = np.dot(np.linalg.pinv(matrix), vector)
+
+        # Compute the `batch_size` to use up to `max_memory`
+        optimal_batch_size = (max_memory - model_size) / item_size
+        optimal_batch_size = int(optimal_batch_size)
```
HollowPrincess, line 287:
```suggestion
        optimal_batch_size = (max_memory - model_size) // item_size
```
If you wanted floor rounding

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,336 @@
...
+
+        return {'batch_size': optimal_batch_size,
+                'item_size': item_size,
+                'model_size': model_size,
+                'table': table}
+
+    def get_memory_utilization(self, batch_size, method='train', inputs=None, targets=None, n=20, frequency=0.05,
+                               time_threshold=3, tail_size=20, std_threshold=0.1):
+        """ For a given `batch_size`, make `inputs` and `targets` and compute memory utilization. """
+        inputs = inputs or self.make_placeholder_data(batch_size, to_device=False) # maybe use to_device=False
```
HollowPrincess, line 297: I can't find the method `make_placeholder_data` in this file.

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # Clear the GPU from potential previous runs
+        self.garbage_collection_cuda()
+        return self._get_memory_utilization(method=method, inputs=inputs, targets=targets, n=n, frequency=frequency,
+                                            time_threshold=time_threshold,
+                                            tail_size=tail_size, std_threshold=std_threshold)
+
+    def _get_memory_utilization(self, method, inputs, targets, n, frequency,
+                                time_threshold, tail_size, std_threshold):
+        """ Run method `n` times and make sure that memory measurements are stable. """
```
HollowPrincess, line 313: Not clear description.

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,336 @@
...
+        with GPUMemoryMonitor(frequency=frequency) as monitor:
+            for _ in range(n):
+                if method == 'train':
+                    _ = self.train(inputs=inputs, targets=targets, microbatch_size=False)
+                elif method == 'predict':
+                    _ = self.predict(inputs=inputs, microbatch_size=False)
+
+        if not self.config.get('benchmark'):
+            data = monitor.data
```
HollowPrincess, line 322: Should be placed before `if`

HollowPrincess (changes requested):

batchflow/models/torch/base_batchsize_mixin.py
HollowPrincess: If it is should be defined in the child class, please, provide an empty method with such a comment.

batchflow/models/torch/base_batchsize_mixin.py
BulatVakhitov: In your implementation optimal_batch_size will be float, because both numenator and denominator are floats. So anyway I have to convert it to int.

batchflow/models/torch/base_batchsize_mixin.py
BulatVakhitov: `start_batch_size` is an integer, so I can't use `list(start_batch_size)`, because it takes only iterable objects.

batchflow/models/torch/base_batchsize_mixin.py
BulatVakhitov: The `make_placeholder_data` method lies in `TorchModel` class. The main idea of mixins to the `TorchModel` class is to split one huge class into several modules for convenience. So following the example of other mixins to this class, I used child's method in a mixin's method.

batchflow/models/torch/base_batchsize_mixin.py
HollowPrincess: You are right, my mistake

batchflow/models/torch/base_batchsize_mixin.py
```diff
@@ -0,0 +1,332 @@
...
+
+        if update_method == 'binary':
+            n_iters = int(np.ceil(np.log2(high - low)))
+            generator = self._binary_batch_size_generator(low=low, high=high, max_memory=max_memory)
+
+        elif update_method == 'bruteforce':
+            n_iters = None
+            generator = self._bruteforce_batch_size_generator(factor=factor, max_memory=max_memory)
+        else:
+            raise ValueError("Wrong update method! Could be `bruteforce` or `binary`")
```
SergeyTsimfer, line 159:
```
Unknown `update_method`: select either `'predictive'` or `'binary'`.
```

SergeyTsimfer (approved):

PR #756, opened 2024-07-11
Monitor and pipeline improvements

SergeyTsimfer:
This PR brings incremental improvements to `Pipeline` and `Monitor` classes:

- for `Pipeline`, this PR fixes a long standing bug of finishing the main thread after getting the end of pipeline signal, but before processing all the batches. Other than that, it implements a way to provide a custom `Executor` instance as a target for `prefetch`: this helps a lot to mitigate the lack of `cudnn.benchmark` cache, which forces each new thread to re-benchmark the model;
- for `Monitors`, it gives a better distinction between system-wide (and rarely used) resources and the ones belonging to a process; also, it gives the ability to monitor other process by providing its PID. 
Finally, this PR implements serialization protocols for `process` and `GPU` monitors.


AlexeyKozhevin (approved):

AlexeyKozhevin:
Please update with changes from `master` branch.

PR #757, opened 2024-07-15
Convert outputs to dict

BulatVakhitov:
This PR makes it possible to work with outputs as dicts. The main points are:

- The idea behind PR is to get rid of memorising that sigmoid output is on index 0, predictions is on index 1, and loss is on index 2. Now you can address them by their name.
- In case the model has multiple predictions, it's now user's responsibility to parse these predictions.
- You still can pass outputs as list, string or callable, but under the hood they all are converted to dict. The resulting processed output would be dict if outputs is a dict, list if outputs is a list and tensor otherwise.


batchflow/models/torch/base.py
```diff
@@ -1010,10 +991,17 @@ def train(self, inputs, targets, outputs=None, mode='train', lock=True, profile=
...
-            Pre-defined tensors are `predictions`, `loss`, and `predictions_{i}` for multi-output models.
+            Each string defines a tensor to get and should be one of:
+                - pre-defined tensors, which are `predictions`, `loss`, and `predictions_{i}` for multi-output models.
+                - pre-defined operations, which are `softplus`, `sigmoid`, `sigmoid_uint8`, `sigmoid_int16`, 
+                `proba`, `labels`. Work only with len(predictions) == 1
+                - layer id, which describes how to access the layer through a series of `getattr` and `getitem` calls.
+                Allows to get intermediate activations of a neural network.
+            Each callable defines a function that should be applied to predictions.
+            If outputs are dict, then keys are strings and they are considered as output_names. The values should be
+            either callables, pre-defined tensors, pre-defined operations or layer id.
```
SergeyTsimfer, line 1006:
```
A tensor to get and should be one of:
- a string, which can be ...
- - ...
- - ...
- a callable, ...
- a sequence, where each item is one of the previous types. Result of this method is guaranteed to have the same order of elements;
- a dict, where each value is one of the previous types. Result of this method is a dictionary with the same keys and requested tensors as values.
```

batchflow/models/torch/base.py
```diff
@@ -1245,20 +1236,24 @@ def _train(self, inputs, targets, outputs, sync_frequency, sam_rho, sam_individu
...
+        output_container = self.compute_outputs(predictions=predictions, operations=outputs_dict)
+        if 'predictions' in outputs_dict.values():
+            # in case there are multiple output_names with the same operation == `predictions`. Same for `loss`
+            predictions_names_list = [output_name for output_name, operation in outputs_dict.items() \
+                                      if operation == 'predictions']
+            output_container.update({output_name: predictions for output_name in predictions_names_list})
+        if 'loss' in outputs_dict.values():
+            losses_names_list = [output_name for output_name, operation in outputs_dict.items() \
+                                 if operation == 'loss']
+            output_container.update({output_name: loss for output_name in losses_names_list})
```
SergeyTsimfer, line 1250:
```python
for output_name, requested in output_dict.items():
    if requested == 'prediction':
        output_container[output_name] = predictions
    elif requested == 'loss':
        output_container[output_name] = loss
```
IMO, looks much simpler. Can be further reduced to 4 lines.

batchflow/models/torch/base.py
```diff
@@ -1406,13 +1404,20 @@ def predict(self, inputs, targets=None, outputs=None, lock=True, microbatch_size
...
             # Aggregate the outputs from microbatches
-            result = self.aggregate_microbatches(outputs, chunked_outputs, chunk_sizes, single_output)
+            result = self.aggregate_microbatches(outputs_dict=outputs_dict, chunked_outputs=chunked_outputs,
+                                                 chunk_sizes=chunk_sizes)
+
+            # convert results to tensor or list depending on outputs type. If dict, do nothing
+            if isinstance(outputs, str) or callable(outputs):
+                result = list(result.values())[0]
+            elif isinstance(outputs, list):
+                result = list(result.values())
```
SergeyTsimfer, line 1422: what about `tuple` / `set`?

batchflow/models/torch/base.py
```diff
@@ -1555,110 +1562,94 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
-        if single_output:
-            result = result[0]
+                    result[output_name] = torch.mean(torch.stack(chunked_output))
         return result
 
 
-    def compute_outputs(self, predictions):
-        """ Produce additional outputs, defined in the config, from `predictions`.
-        Also adds a number of aliases to predicted tensors.
+    def compute_outputs(self, predictions, operations):
```
SergeyTsimfer, line 1574: Probably, can move code for adding `predictions` / `loss` here

batchflow/models/torch/base.py
```diff
@@ -1555,110 +1562,94 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
-                name = ''
+                result = predictions
             elif callable(operation):
-                result = operation(tensor)
-                name = operation.__name__
+                result = operation(predictions)
+            elif isinstance(operation, LayerHook):
+                operation.close()
+                result = operation.activation
+            elif isinstance(operation, str) and re.match(r"predictions_[0-9]+", operation):
```
SergeyTsimfer, line 1595: either don't use regexp here or compile it once at the module / class level

batchflow/models/torch/base.py
```diff
@@ -1555,110 +1562,94 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
+                i = int(re.findall(r"\d+", operation)[0]) # get the ordinal number of prediction
+                result = predictions[i]
             else:
+                if isinstance(predictions, (tuple, list)) and not len(predictions) == 1:
+                    raise ValueError('Default operations can`t be applicable to multi output predictions.')
```
SergeyTsimfer, line 1600: `can't be applied`, maybe?

batchflow/models/torch/base.py
```diff
@@ -1555,110 +1562,94 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
-                name = operation
-        return result, name
+        return result
 
 
     def prepare_outputs(self, outputs):
-        """ Add the hooks to all outputs that look like a layer id. """
-        result = []
-        for output_name in outputs:
+        """ Add the hooks to all outputs that look like a layer id. Also convert outputs to dict"""
```
SergeyTsimfer, line 1621: Formatting

batchflow/models/torch/base.py
```diff
@@ -1555,110 +1562,94 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
+        result = OrderedDict()
+
+        if not isinstance(outputs, dict): # then outputs should be str, callable, list or tuple
+            outputs = list(outputs) if isinstance(outputs, (list, tuple)) else [outputs]
+            processed_outputs = OrderedDict()
+            for output in outputs:
+                if isinstance(output, str):
+                    processed_outputs[output] = output
+                elif callable(output):
+                    processed_outputs[output.__name__] = output
```
SergeyTsimfer, line 1631: Why is that needed? callables are hashable

batchflow/models/torch/base.py
```diff
@@ -1555,112 +1542,103 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
-        if single_output:
-            result = result[0]
+                    result[output_name] = torch.mean(torch.stack(chunked_output))
         return result
 
 
-    def compute_outputs(self, predictions):
-        """ Produce additional outputs, defined in the config, from `predictions`.
-        Also adds a number of aliases to predicted tensors.
+    def compute_outputs(self, predictions, operations, targets=None, loss=None):
```
SergeyTsimfer, line 1554: I would say that renaming `operations` to `requested_outputs` would make things much easier to understand. Or maybe to `outputs_dict`, to keep in line with the rest of the code

batchflow/models/torch/base.py
```diff
@@ -1555,112 +1542,103 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
+        for name, operation in operations.items():
+            if operation == 'predictions':
+                result[name] = predictions
+            elif operation == 'loss':
+                if loss is not None: # the loss is precomputed, so just store it
+                    result[name] = loss
+                elif targets is not None:
+                    targets = self.transfer_to_device(targets)
+                    loss = self.loss(predictions, targets)
+                    result[name] = loss
```
SergeyTsimfer, line 1568: What do you think about keeping this part in the `_predict` method?

batchflow/models/torch/base.py
```diff
@@ -1555,112 +1542,103 @@ def aggregate_microbatches(self, outputs, chunked_outputs, chunk_sizes, single_o
...
-                    result = tensor.argmax(dim=1)
+                    result = predictions.argmax(dim=1)
                 else:
                     raise ValueError(f'Unknown type of operation `{operation}`!')
-                name = operation
-        return result, name
+        return result
 
 
     def prepare_outputs(self, outputs):
```
SergeyTsimfer, line 1614: `prepare_outputs` sounds very much alike to `compute_outputs`, while the intent here is to `prepare user-passed argument "outputs" to a form that we internally use". Do you see any better names?

SergeyTsimfer (commented):
Other than the naming, this PR looks great. If possible, make a better distinction between internal/external variables, and it is ready to merge.

Good job:)

SergeyTsimfer (approved):

AlexeyKozhevin (approved):

PR #759, opened 2024-07-24
fix release

BulatVakhitov:
This PR updates the release workflow to implement the trusted publishers workflow as suggested [here](https://docs.pypi.org/trusted-publishers/using-a-publisher/)


.github/workflows/release-check.yml
```diff
@@ -22,12 +25,12 @@ jobs:
...
         poetry install
         poetry build
-        poetry config repositories.testpypi https://test.pypi.org/legacy/
-        poetry config pypi-token.testpypi $_PASSWORD
-        poetry publish --repository testpypi
+
+    - name: Publish
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        repository-url: https://test.pypi.org/legacy/
```
roman-kh, line 36: The last line in the file should be empty

.github/workflows/release.yml
```diff
@@ -42,11 +45,10 @@ jobs:
...
-        _PASSWORD: ${{ secrets.TEST_PYPI_TOKEN }}
+    - name: Build
       run: |
         poetry install
         poetry build
-        poetry config pypi-token.pypi $_PASSWORD
-        poetry publish
+
+    - name: Publish
+      uses: pypa/gh-action-pypi-publish@release/v1
```
roman-kh, line 54: The last line in the file should be empty

roman-kh (changes requested):

AlexeyKozhevin (approved):

PR #760, opened 2024-08-01
Improvements

SergeyTsimfer:
This PR:
- adds a new `IF` named expression, which uses a `condition` to select outcome.
This greatly simplifies some of the named expressions, that we've used for selecting, say, model train mode or shapes (looking at you, `(I.current < 500) * 'train' + (I.current >=500) * 'eval'`)
- adds a way to use callbacks in `Notifier`.
Again, simplifies the interface: instead of subclassing the `Notifier` or `DummyBar` or `tqdm.standard` for adding an action (for example, logging to a custom state-recorder), one can use a simple function to do the same.
- grants a finer control over the model `init_weights` and `mode` parameter of `train/predict` methods.
The latter can be used for implementing custom schemes like `eval mode but with the dropout turned on`, which is somewhat cumbersome to implement without the improvements provided in the PR.

### DIFF
--- a/batchflow/named_expr.py
+++ b/batchflow/named_expr.py
@@ -343,12 +343,15 @@ class AlgebraicNamedExpression(NamedExpression):
             a = eval_expr(self.a, _call=False, **kwargs)
         else:
             a = eval_expr(self.a, **kwargs)
-        b = eval_expr(self.b, **kwargs)
-        c = eval_expr(self.c, **kwargs)
+
         if self.op in UNARY_OPS:
             return OPERATIONS[self.op](a)
+
+        b = eval_expr(self.b, **kwargs)
         if self.op in BINARY_OPS:
             return OPERATIONS[self.op](a, b)
+
+        c = eval_expr(self.c, **kwargs)
         return OPERATIONS[self.op](a, b, c)
 
     def assign(self, value, **kwargs):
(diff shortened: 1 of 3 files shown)


batchflow/named_expr.py
```diff
@@ -401,6 +404,58 @@ def __repr__(self):
         return 'Unknown expression'
 
 
+class IF(NamedExpression):
+    """ Select either ``true`` or ``false``, based on ``condition``.
+    Useful for simple variables that change along the run of a pipeline.
+
+    Examples
+    --------
+    Select model mode based on the current pipeline iteration::
```
HollowPrincess, line 413: Something is not OK with amount of symbols in the entire docstring: too much double `` and ::.

batchflow/named_expr.py
```diff
@@ -401,6 +404,58 @@ def __repr__(self):
         return 'Unknown expression'
 
 
+class IF(NamedExpression):
+    """ Select either ``true`` or ``false``, based on ``condition``.
+    Useful for simple variables that change along the run of a pipeline.
+
+    Examples
```
HollowPrincess, line 411: Not as in the numpy docstyle 
https://numpydoc.readthedocs.io/en/latest/format.html#examples

batchflow/named_expr.py
```diff
@@ -401,6 +404,58 @@ def __repr__(self):
...
+    Train the last 20% with larger batch size::
+        batch_size = IF(condition=I.ratio > 0.8, true=256, false=128)
+
+    Notes
+    -----
+    An alternative to this named expression is to use ``F``::
+        def select_batch_size(ratio):
+            return 256 if ratio > 0.8 else 128
+
+        batch_size = F(select_batch_size)(ratio=I.ratio)
```
HollowPrincess, line 425: You can see an example of multiple lines code in the docs here: https://numpydoc.readthedocs.io/en/latest/format.html#examples

batchflow/named_expr.py
HollowPrincess: And here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

batchflow/notifier.py
```diff
@@ -430,7 +431,8 @@ def update_plot(self, index=0, add_suptitle=False, savepath=None, clear_display=
             self.plotter.config['suptitle'] = self.bar.format_meter(**fmt)
             self.plotter.annotate()
 
-        for i, container in enumerate(self.data_containers):
+        data_containers = [container for container in self.data_containers if container['data'] is not None]
+        for i, container in enumerate(data_containers):
             if i >= index:
                 subplot_index = i - index
                 subplot_config = plot_config.maybe_index(subplot_index)
```
HollowPrincess, line 438: Why you use comprehension and cycle instead of skipping wrong cntainer in cycle? Looks like double itration.
Proposal:
```Python
for i, container in enumerate(self.data_containers):
    if container['data'] is None:
        continue
```

batchflow/named_expr.py
SergeyTsimfer: Pretty much the entire file is not in numpy docstyle, as well as Pipeline and some other files. [Local consistency is the most important](https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds)

batchflow/notifier.py
SergeyTsimfer: Skipping the items in the same loop would make it so that `i` variable is also skipped (and, therefore, incorrect)

PR #774, opened 2025-01-28
Fix save model

AlexeyKozhevin:
Add `ignore_attributes` to `save` method to avoid serialization issues for optimizer with the newest versions of pytorch and python


batchflow/models/torch/base.py
```diff
@@ -1719,7 +1726,9 @@ def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_op
             torch.onnx.export(self.model.eval(), inputs, path_onnx, opset_version=opset_version)
 
             # Save the rest of parameters
-            preserved = set(self.PRESERVE) - set(['model', 'loss', 'optimizer', 'scaler', 'decay'])
+            ignore_attributes = set(ignore_attributes) + set(['model', 'loss', 'optimizer', 'scaler', 'decay'])
```
SergeyTsimfer, line 1729: should be moved up, where the `ignore_attributes` is parsed

PR #784, opened 2025-06-09
Add save in safetensors format

igor-iusupov:


AlexeyKozhevin (changes requested):

batchflow/models/torch/base.py
```diff
@@ -1751,11 +1754,23 @@ def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_op
...
+        elif use_safetensors:
+            from safetensors.torch import save_file
+            state_dict = self.model.state_dict()
+
+            path_safetensors = path_safetensors or (path + "safetensors")
+            save_file(state_dict, path_safetensors)
+
+            if pickle_metadata:
+                torch.save({'safetensors': True, 'path_safetensors': path_safetensors, **preserved_dict},
+                        path, pickle_module=pickle_module, **kwargs)
```
AlexeyKozhevin, line 1773: There is no definition of `preserved_dict `

AlexeyKozhevin (changes requested):

AlexeyKozhevin:
Try to run `save` and `load` in different scenarios. Thus, you will find some erros, e.g. missing `preserved_dict` for safetensors case

batchflow/models/torch/base.py
```diff
@@ -1828,6 +1887,11 @@ def load(self, file, make_infrastructure=False, mode='eval', pickle_module=dill,
                 self._loaded_from_onnx = True
                 self.disable_training = True
 
+            if "safetensors" in checkpoint:
+                from safetensors.torch import load_file
+                state_dict = load_file(checkpoint['path_safetensors'], device=device)
+                self.model.load_state_dict(state_dict)
```
AlexeyKozhevin, line 1893: Did you test `load` method? As i understand, `self.model` is `None` before that line and that line will raise Exception

AlexeyKozhevin (changes requested):

AlexeyKozhevin:
Fix linting issues

batchflow/models/torch/base.py
```diff
@@ -1816,7 +1816,10 @@ def load(self, file, make_infrastructure=False, mode='eval', pickle_module=dill,
             if file.endswith(".safetensors"):
                 from safetensors.torch import load_file
                 state_dict = load_file(file, device=device)
-                self.model.load_state_dict(state_dict)
+
+                with torch.no_grad():
```
AlexeyKozhevin, line 1820: What if one need to load model for the following training?

AlexeyKozhevin (changes requested):

batchflow/models/torch/base.py
```diff
@@ -1751,11 +1755,26 @@ def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_op
...
+                preserved_dict = {item: getattr(self, item) for item in preserved}
+                torch.save({'openvino': True, 'path_openvino': path_openvino, **preserved_dict},
+                        path, pickle_module=pickle_module, **kwargs)
+
+        elif use_safetensors:
+            from safetensors.torch import save_file
+            state_dict = self.model.state_dict()
+
+            path_safetensors = path_safetensors or (path + '.safetensors')
+            save_file(state_dict, path_safetensors)
```
AlexeyKozhevin, line 1770: In the case when I want to save only safetensors model I can
1) `model.save(path='model.pt', use_safetensors=True, path_safetensors='model.safetensors', pickle_metadata=False)` -> in that case I don't use `path` which is necessary argument and i can't skip it
2) `model.save(path='model.pt', use_safetensors=True, pickle_metadata=False)` -> the model will be in `model.pt.safetensors`, I am not sure that two extensions is a good idea

Think how to avoid it. For example, make `path=None`, by default.

AlexeyKozhevin (changes requested):

AlexeyKozhevin (changes requested):

batchflow/models/torch/base.py
```diff
@@ -1709,14 +1708,11 @@ def save(self, path=None, use_onnx=False, path_onnx=None, use_openvino=False, pa
...
-
-        if path is None:
-            dirname = os.getcwd()
-            path = os.path.join(dirname, "model.pt")
-        else:
-            dirname = os.path.dirname(path)
-            if dirname and not os.path.exists(dirname):
-                os.makedirs(dirname)
+        available_formats = ("pt", "onnx", "openvino", "safetensors")
+        assert format in available_formats, f"Format must be in {available_formats}"
```
roman-kh, line 1712: You can use `assert` to check the arguments in internal methods, but in public methods you need to raise `ValueError`

batchflow/models/torch/base.py
```diff
@@ -1669,7 +1669,7 @@ def convert_outputs(self, outputs):
 
 
     # Store model
-    def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_openvino=None,
+    def save(self, path, format="pt", pickle_metadata=False,
```
AlexeyKozhevin, line 1672:
```suggestion
    def save(self, path, fmt="pt", pickle_metadata=True,
```

batchflow/models/torch/base.py
```diff
@@ -1720,25 +1727,26 @@ def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_op
...
         ignore_attributes = set(ignore_attributes)
 
-        if use_onnx:
+        if format == "onnx":
             if batch_size is None:
                 raise ValueError('Specify valid `batch_size`, used for model inference!')
 
             inputs = self.make_placeholder_data(batch_size=batch_size, unwrap=False)
-            path_onnx = path_onnx or (path + '_onnx')
+            path_onnx = path if path.endswith(".onnx") else os.path.join(dirname, "model.onnx")
```
AlexeyKozhevin, line 1735: There is no way to call `model.save('my_model.abcd', format="onnx", pickle_metadata=False)` and get a model by path `'my_model.abcd'` which user specified. User must be able to choose extension which he want.

So 
1) use `path` as `path_onnx` if `pickle_metadata=False`,
2) use `os.path.splitext(path)[0]` + '.onnx` if `pickle_metadata=True`.

In that logic `path` is a path to the main file which is dumped. If `pickle_metadata=False`, it is the model itself, if `pickle_metadata=True`, it is pickled file.

The same for other formats.

batchflow/models/torch/base.py
```diff
@@ -1689,6 +1689,8 @@ def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_op
             If provided, then path to store the ONNX model; default `path_onnx` is `path` with '_onnx' postfix.
         use_openvino: bool
             Whether to store model as openvino xml file.
+        use_safetensors: bool
+            Whether to store model as safetensors file.
```
AlexeyKozhevin, line 1693: Update docsting

batchflow/models/torch/base.py
AlexeyKozhevin: `format` is a name of a standard function in Python, try to avoid such names for variables if it is possible (like `slice`, `max`, etc)

batchflow/models/torch/base.py
```diff
@@ -1669,7 +1669,7 @@ def convert_outputs(self, outputs):
 
 
     # Store model
-    def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_openvino=None,
+    def save(self, path, fmt="pt", pickle_metadata=False,
```
AlexeyKozhevin, line 1672: To keep current default behaviour:

```suggestion
    def save(self, path, fmt="pt", pickle_metadata=True,
```

batchflow/models/torch/base.py
```diff
@@ -1682,17 +1682,8 @@ def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_op
...
-            Used only if `use_onnx` is True.
-            If provided, then path to store the ONNX model; default `path_onnx` is `path` with '_onnx' postfix.
-        use_openvino: bool
-            Whether to store model as openvino xml file.
-        path_openvino : str, optional
-            Used only if `use_openvino` is True.
-            If provided, then path to store the openvino model; default `path_openvino` is `path` with '_openvino'
-            postfix.
+        fmt: str
+            Weights format. Available formats: "pt", "onnx", "openvino", "safetensors"
```
AlexeyKozhevin, line 1686: Add `pickle_metadata`

AlexeyKozhevin (changes requested):

batchflow/models/torch/base.py
```diff
@@ -1793,6 +1810,50 @@ def load(self, file, make_infrastructure=False, mode='eval', pickle_module=dill,
         else:
             self._parse_devices()
 
+        if isinstance(file, str):
+            if fmt == "safetensors" or (fmt is None and file.endswith(".safetensors")):
+                from safetensors.torch import load_file
+                state_dict = load_file(file)
+
+                self.initialize()
```
AlexeyKozhevin, line 1818: Why do you need it here? If you try to load a safetensors model using your function, you will get an error:
```model.save('tmp.safetensors', fmt='safetensors', pickle_metadata=False)``` -> `ValueError: Set "loss" in model configuration!`

File: batchflow/models/torch/base.py [hunk 12]
```diff
@@ -1669,7 +1669,7 @@ def convert_outputs(self, outputs):
 
 
     # Store model
-    def save(self, path, use_onnx=False, path_onnx=None, use_openvino=False, path_openvino=None,
+    def save(self, path, fmt="pt", pickle_metadata=True,
```
AlexeyKozhevin, line 1672: Make `fmt=None` by default to get it from an extension in `path`

AlexeyKozhevin (changes requested):

AlexeyKozhevin (changes requested):

